PR #2, opened 2020-06-25
Add $(HELPER_DIR) and others to Makefile.common

karkhaz:
This commit:

- Adds the $(HELPER_DIR) variable, set to the parent of the 'proofs'
  directory, to Makefile.common
- Renames the "sources" directory to "models", so that it doesn't get
  confused with $(SRCDIR)
- Adds the $(MODELS_DIR) and $(STUBS_DIR) variables for easy reference
  from proof makefiles
- Adds a recommendation that project owners should set $(INCLUDES) to
  contain "-I$(HELPER_DIR)/include". This commit does not actually set
  this variable, since the decision of which order it comes in in the
  include list is a matter of project policy.

### DIFF
--- a/template-for-repository/proofs/Makefile-project-defines
+++ b/template-for-repository/proofs/Makefile-project-defines
@@ -20,6 +20,11 @@
 # LINK_FLAGS =
 
 # Preprocessor include paths -I...
+# Consider adding
+#     INCLUDES += -I$(CBMC_ROOT)/include
+# You will want to decide what order that comes in relative to the other
+# include directories in your project.
+#
 # INCLUDES =
 
 # Preprocessor definitions -D...
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -83,6 +83,9 @@ default: report
 # Absolute path to the directory containing this Makefile.common
 # See https://ftp.gnu.org/old-gnu/Manuals/make-3.80/html_node/make_17.html
 PROOF_ROOT = $(dir $(abspath $(filter %/Makefile.common,$(MAKEFILE_LIST))))
+CBMC_ROOT = $(shell dirname $(PROOF_ROOT))
+PROOF_STUB = $(CBMC_ROOT)/stubs
+PROOF_SOURCE = $(CBMC_ROOT)/sources
 
 # Project-specific definitions to override default definitions below
 #   * Makefile-project-defines will never be overwritten
--- a/template-for-repository/sources/README.md
+++ b/template-for-repository/sources/README.md
@@ -1,6 +1,6 @@
 CBMC proof source code
 ======================
 
-This directory contains source code written for CBMC proof.  It is
+This directory contains source code written for CBMC proofs.  It is
 common to write some code to model aspects of the system under test,
 and this code goes here.


template-for-repository/proofs/Makefile-project-defines
```diff
@@ -20,6 +20,11 @@
 # LINK_FLAGS =
 
 # Preprocessor include paths -I...
+# It is likely that you will want to add at least
+#     INCLUDES += -I$(HELPER_DIR)/include
+# but you will want to decide what order that comes in relative to the
+# other includes in your project.
+#
```
markrtuttle, line 27: I'd shorten this to "Consider adding INCLUDES += ..."

template-for-repository/models/README.md
```diff
@@ -1,6 +1,6 @@
 CBMC proof source code
```
markrtuttle, line 1: This renaming of sources to models is a bit gratuitous since the MQTT and HTTP folks have already started populating sources and includes with project code.  If it is likely people will think cbmc/sources is part of the source tree, it is equally likely that people will think cbmc/include is part of the source tree.  It might be worth using the names PROOF_SOURCE=cbmc/sources and PROOF_INCLUDE=cbmc/include in the Makefile.common.

template-for-repository/proofs/Makefile.common
```diff
@@ -83,6 +83,9 @@ default: report
 # Absolute path to the directory containing this Makefile.common
 # See https://ftp.gnu.org/old-gnu/Manuals/make-3.80/html_node/make_17.html
 PROOF_ROOT = $(dir $(abspath $(filter %/Makefile.common,$(MAKEFILE_LIST))))
+HELPER_DIR = $(shell dirname $(PROOF_ROOT))
+STUBS_DIR = $(HELPER_DIR)/stubs
+MODELS_DIR = $(HELPER_DIR)/models
```
markrtuttle, line 88: Again, I'd use PROOF_STUB, PROOF_SOURCE, PROOF_INCLUDE.

HELPER_DIR is a name we commonly use, but it is pretty meaningless, and I intentionally did not use it.  I'd rather something like CBMC_ROOT for cbmc, and have PROOF_STUB, PROOF_SOURCE, PROOF_INCLUDE be the stubs, sources, include subdirs.

template-for-repository/proofs/Makefile-project-defines
karkhaz: Done

template-for-repository/proofs/Makefile.common
karkhaz: I changed HELPER_DIR to CBMC_ROOT.
I changed STUBS_DIR to PROOF_STUB.
I changed MODELS_DIR to PROOF_SOURCE and renamed "models" back to "sources".

I did not add PROOF_INCLUDE because we don't know what the include order will be. The recommendation in the `Makefile-project-defines` file is now to add `-I$(CBMC_ROOT)/include` to the `INCLUDES` variable. I could set `PROOF_INCLUDE = $(CBMC_ROOT)/include` and recommend to add `PROOF_INCLUDE` to `INCLUDE`, but it seems unnecessary to create a variable that will only ever be used once (we will never append to `PROOF_INCLUDE` and never use it in any other context).

template-for-repository/models/README.md
karkhaz: Done

PR #4, opened 2020-06-26
Copy rather than symlink user-modifiable files

karkhaz:
This commit makes the setup script copy the following files into the
project repository, rather than symbolically linking them:

  * Makefile-project-defines
  * Makefile-project-targets
  * Makefile-project-testing

This is because project owners are expected to modify these files in the
project repository. These changes should not be written to the
submodule, because the submodule would then become dirty and impossible
to update.

### DIFF
--- a/scripts/util.py
+++ b/scripts/util.py
@@ -5,11 +5,22 @@
 
 import logging
 import os
+import shutil
 
 REPOSITORY_TEMPLATES = "template-for-repository"
 PROOF_TEMPLATES = "template-for-proof"
 
 PROOF_DIR = "proofs"
+
+# There are some files that we should copy to the project repository rather than
+# symlinking. This is because users are expected to modify these files. If the
+# files were symlinks, then modifying them would dirty up this submodule, which
+# would prevent project owners from cleanly updating it.
+COPY_INSTEAD = [
+    "Makefile-project-defines",
+    "Makefile-project-targets",
+    "Makefile-project-testing",
+]
 
 ################################################################
 
@@ -64,6 +75,7 @@ def files_under_root(root):
     finally:
         os.chdir(cwd)
 
+
 def link_files(name, src, dst):
     """Link file dst/name to file src/name, return number skipped"""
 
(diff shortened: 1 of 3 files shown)


markrtuttle:
I've put a lot of thought into this.  It is not at all clear to me that copy is the right thing to do.  For example, we deployed the templates and everything worked until we tried to run CI.  Then we realized the code to generate the cbmc-batch.yaml was missing, and it was a trivial matter to update Makefile-project-testing and everyone got the code with the next submodule update.  This is a very slippery slope here.  For example, you could argue that the REAMDE.md files should be updated by the project to describe project details.  What is the criterion for making the copy/symlink decision?  My final decision was to symlink things by default.  Projects who DO actually update one of these files can delete the symlink and edit the file in place, and know that the setup.py script will never overwrite their work.  I am not in favor of this change.

karkhaz:
> Projects who DO actually update one of these files can delete the symlink and edit the file in place, and know that the setup.py script will never overwrite their work

If they delete the symlink & edit the file in place, then they will not receive any update when you do this:

> it was a trivial matter to update Makefile-project-testing and everyone got the code with the next submodule update

I believe the correct solution is that the three files above should be copied. In addition, there should be three corresponding files (`Makefile-template-*` rather than `Makefile-project-*`) that get symlinked. Any update you want everybody to get goes into the latter three files, which contain a prominent comment saying that they are not to be modified. I propose that I add this to the PR.

> For example, you could argue that the REAMDE.md files should be updated by the project to describe project details

There are two cases:
* Makefiles: this is easy. If we expect people to update, copy. If we want to push updates, symlink. If we want both, do both, as I mentioned above. It wouldn't surprise me if somebody wanted to add some extra rules to Makefile.common one day. If that day ever comes, you could just add `sinclude Makefile.common.local` to `Makefile.common` and push that change.
* READMEs and other prose. This one is a little more annoying because you can't include READMEs into other READMEs. This runs the risk of people modifying a README locally, but then you change the README to have some new instructions or documentation, but the project doesn't get those changes because they aren't copied. I don't know that there's an elegant solution here, but I also don't think it's a big deal. I would err on the side of copying the readme and letting people change it. If it's vital for people to update their readme with some new instructions, you can just email them and tell them to merge the changes in.

In summary: I propose copying the three files I'm copying, plus the readme, and also add three `Makefile-project-*` files that get symlinked for you to put updates into.

karkhaz:
Addressed our offline conversation. I've moved the contents of Makefile-project-testing into the bottom of Makefile.common and renamed a variable.

scripts/util.py
```diff
@@ -5,12 +5,23 @@
...
 
 REPOSITORY_TEMPLATES = "template-for-repository"
 PROOF_TEMPLATES = "template-for-proof"
 
 PROOF_DIR = "proofs"
 
+# There are some files that we should copy to the project repository rather than
+# symlinking. This is because users are expected to modify these files. If the
+# files were symlinks, then modifying them would dirty up this submodule, which
+# would prevent project owners from cleanly updating it.
```
markrtuttle, line 18: It is enough to say "File to copy instead of symlinking."

PR #5, opened 2020-07-05
Add a .gitignore to the proof directory

karkhaz:
The common Makefile writes temporary files into the `gotos`, `logs`,
`html`, and `report` subdirectories of each proof directory by default.
This commit adds a .gitignore to be symlinked into the top-level proof
directory that ignores the contents of these directories, as
proof-writers will almost certainly want to have this file anyway.


karkhaz:
Addressed offline comments by @markrtuttle

PR #7, opened 2020-07-06
Make all 3 CBMC jobs run in parallel

karkhaz:
This commit adds the `-j4` flag to the list of command line arguments
to Make, so that all three runs of CBMC can run in parallel on a
particular proof.

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -65,6 +65,9 @@ SHELL=/bin/bash
 
 default: report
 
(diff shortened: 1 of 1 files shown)


template-for-repository/proofs/Makefile.common
```diff
@@ -65,6 +65,9 @@ SHELL=/bin/bash
 
 default: report
 
+# Run all CBMC jobs in parallel for a particular proof
+MAKEFLAGS += -j4
```
danielsn, line 69: If someone passed a different `-j` on the commandline, which one wins?

PR #8, opened 2020-07-10
Add license header for prepare.py

karkhaz:
Also fix whitespace in prepare.py.


markrtuttle-docker:
This is the long header.  Please use the sort version of the license used in all the other files.

PR #12, opened 2020-07-15
Add arpa-related content in Makefile.common

ArenBabikian:
I have added `arpa`-related content into the [Makefile.common](https://github.com/awslabs/aws-templates-for-cbmc-proofs/blob/master/template-for-repository/proofs/Makefile.common).

The proposed additions do not necessarily modify the current user experience for writing and running proofs (in the case where users do not want to use `arpa`. Furthermore, with my proposed additions, integrating `arpa` into existing projects will be as simple as following these two steps:
1. Submodule [aws-proof-build-assistant](https://github.com/awslabs/aws-proof-build-assistant) into the [`CBMC_ROOT`](https://github.com/awslabs/aws-templates-for-cbmc-proofs/blob/afeb589f31e112ccf615f2baac650d13197b2346/template-for-repository/proofs/Makefile.common#L86) directory of the project.
2. Add project-specific `cmake` flags to [`Makefile-project-defines`](https://github.com/awslabs/aws-templates-for-cbmc-proofs/blob/master/template-for-repository/proofs/Makefile-project-defines)


template-for-repository/proofs/Makefile.common
```diff
@@ -222,6 +222,8 @@ GOTO_INSTRUMENT ?= goto-instrument
 VIEWER ?= cbmc-viewer
 MAKE_SOURCE ?= make-source
 VIEWER2 ?= cbmc-viewer
+CMAKE ?= cmake
+ARPA ?= $(CBMC_ROOT)/aws-proof-build-assistant/arpa
```
karkhaz, line 226: I think this should be in Makefile-project-defines. You will still be able to use `CBMC_ROOT` because that variable is defined before the local Makefiles are included.

I suggest that you delete this line, and instead write a comment inside Makefile-project-defines suggesting that people define `ARPA`. Something like
```
# Path to Arpa executable
# ARPA = 

```

template-for-repository/proofs/Makefile.common
```diff
@@ -339,6 +344,18 @@ $(HARNESS_GOTO).goto: $(HARNESS_GOTO)4.goto
 goto: $(HARNESS_GOTO).goto
 	echo $(DEPENDENT_GOTOS)
 
+################################################################
+# Targets to run Arpa
+
+arpa_cmake_project:
```
karkhaz, line 350: I think this target should actually be called `$(ARPA_BLDDIR)`, since that is what will be created

template-for-repository/proofs/Makefile.common
```diff
@@ -339,6 +344,18 @@ $(HARNESS_GOTO).goto: $(HARNESS_GOTO)4.goto
...
+	$(CMAKE) $(ARPA_CMAKE_FLAGS) \
+	-DCMAKE_EXPORT_COMPILE_COMMANDS=1 \
+	-B $(ARPA_BLDDIR) \
+	-S $(SRCDIR)
```
karkhaz, line 354: These three lines should be indented with a tab + 2 spaces please.

template-for-repository/proofs/Makefile.common
karkhaz: I guess you want the project owner to set `ARPA_CMAKE_FLAGS` to be the same as the cmake flags used for building the project?

If so, please add a comment to Makefile-project-defines to tell people to set this variable, similar to the `ARPA` comment I mentioned above

template-for-repository/proofs/Makefile.common
```diff
@@ -339,6 +344,18 @@ $(HARNESS_GOTO).goto: $(HARNESS_GOTO)4.goto
...
+
+arpa: arpa_cmake_project
```
karkhaz, line 356: `arpa` should be added to the `PHONY` target please

template-for-repository/proofs/Makefile.common
```diff
@@ -395,6 +412,8 @@ clean:
 	-$(RM) $(DEPENDENT_GOTOS)
 	-$(RM) TAGS*
 	-$(RM) *~ \#*
+	-$(RM) Makefile.arpa
+	-$(RM) -r $(ARPA_BLDDIR)
```
karkhaz, line 416: I wonder whether these two should actually be in `veryclean`. @markrtuttle ?

template-for-repository/proofs/Makefile.common
ArenBabikian: Should I replace line 226 by something like this?
```
ARPA ?= arpa
```
For the case where people forget to define `ARPA` in their local Makefile. This would probably give a more understandable error message too.

karkhaz: That's a good idea. I would go even further:
```
ARPA ?= @echo "You must set `ARPA` in Makefile-project-defines to run arpa in this project"; false
```
😃

template-for-repository/proofs/Makefile-project-defines
```diff
@@ -29,3 +29,9 @@
 
 # Preprocessor definitions -D...
 # DEFINES =
+
+# Path to arpa executable
+# ARPA =
+
+# Flags to pass to cmake for building the project
+# ARPA_CMAKE_FLAGS =
```
karkhaz, line 37: Please ensure files have EOL at the end

karkhaz:
LGTM apart from newline at end-of-file, but let's wait for a review from @markrtuttle or @danielsn or @feliperodri before merging.

template-for-repository/.gitignore
```diff
@@ -7,5 +7,9 @@ proofs/**/html
 # Emitted by CBMC Viewer
 TAGS-*
 
+# Emitted by Arpa
+arpa_cmake
```
markrtuttle, line 11: Since this is a directory, would `arpa_cmake/` be a more specific pattern that would match only a directory?

template-for-repository/proofs/Makefile-project-defines
```diff
@@ -29,3 +29,9 @@
 
 # Preprocessor definitions -D...
 # DEFINES =
+
+# Path to arpa executable
+# ARPA =
```
markrtuttle, line 34: Why comment out this definition?   Both no definition and an empty definition `ARPA = ` cause `$(ARPA)` to evaluate to an empty string.

template-for-repository/proofs/Makefile-project-defines
markrtuttle, line 37: Again, why comment this out?

template-for-repository/proofs/Makefile.common
```diff
@@ -222,6 +222,8 @@ GOTO_INSTRUMENT ?= goto-instrument
 VIEWER ?= cbmc-viewer
 MAKE_SOURCE ?= make-source
 VIEWER2 ?= cbmc-viewer
+CMAKE ?= cmake
+ARPA ?= @echo "You must set ARPA in Makefile-project-defines to run arpa in this project"; false
```
markrtuttle, line 226: This definition forces the use of arpa.  (The evaluation to false will make the invocation of make fail.) I think this is a mistake.  A project should be able to decide on its own whether it will make use of arpa and the targets defined here.

template-for-repository/proofs/Makefile.common
```diff
@@ -233,6 +235,9 @@ HARNESS_GOTO ?= $(GOTODIR)/$(HARNESS_FILE)
 PROJECT_GOTO ?= $(GOTODIR)/$(PROJECT)
 PROOF_GOTO ?= $(GOTODIR)/$(PROOF)
 
+ARPA_BLDDIR ?= $(PROOFDIR)/arpa_cmake
```
markrtuttle, line 238: I would make `arpa_cmake` a variable that can be overridden.  Also, you probably want to use `$(abspath ...)` to ensure that this is an absolute path, you require that in your invocation of cmake below.

template-for-repository/proofs/Makefile.common
markrtuttle: I advocate putting into veryclean only those things that take a long time to rebuild.  This is a left over from the old days.  But cbmc results and the html report are the kinds of things I would put into veryclean.

template-for-repository/proofs/Makefile.common
```diff
@@ -233,6 +235,9 @@ HARNESS_GOTO ?= $(GOTODIR)/$(HARNESS_FILE)
 PROJECT_GOTO ?= $(GOTODIR)/$(PROJECT)
 PROOF_GOTO ?= $(GOTODIR)/$(PROOF)
 
+ARPA_BLDDIR ?= $(abspath $(PROOFDIR)/arpa_cmake)
```
ArenBabikian, line 238: * I notice that all other directory names in this file have been hardcoded: (`PROOF_SOURCE = $(CBMC_ROOT)/sources`,  `GOTODIR ?= $(PROOFDIR)/gotos`, `...`). As a result, I have not made `arpa_cmake` a variable that can be overridden to stay consistent with the rest of the file. Furthermore, this makes it easier to gitignore the arpa build directory, I believe.
Nevertheless, I can make the adjustment if you prefer.

* I have added `$(abspath ...)`.

PR #16, opened 2020-07-27
Convert starter kit to use Litani

karkhaz:
This pull request is emphatically *not* against the master branch. It creates a new branch, `litani-mainline`, which will exist in parallel with `master` until all projects are using Litani. The plan is:
* Each project will be updated to use Litani, and their template submodule will point to the `litani-mainline` branch of this repository rather than `master`.
* Projects that haven't been updated to Litani yet will continue to have a submodule pointing to `master`.
* Development on this repository should continue on the `master` branch as usual. I will forward-port all commits from `master` to `litani-mainline` so that Litani-using projects can use them too.
* When all projects are using Litani, we'll merge the `litani-mainline` branch into master. The merge should be trivial since the commits should all be there, just in a different order.

This pull request updates `Makefile.common` to invoke CBMC and related tools through the Litani build accumulator, hosted [here](https://github.com/awslabs/aws-build-accumulator/).

### DIFF
--- a/template-for-repository/proofs/Makefile-project-defines
+++ b/template-for-repository/proofs/Makefile-project-defines
@@ -13,6 +13,9 @@
 # Makefile.common.
 ################################################################
 
+# Path to litani executable
+# LITANI =
+
 # Flags to pass to goto-cc for compilation (typically those passed to gcc -c)
 # COMPILE_FLAGS =
 
--- a/template-for-repository/proofs/Makefile-project-defines
+++ b/template-for-repository/proofs/Makefile-project-defines
@@ -13,6 +13,9 @@
 # Makefile.common.
 ################################################################
 
+# Name of this proof project, displayed in proof reports
+# PROJECT_NAME =
+
 # Path to litani executable
 # LITANI =
 
--- a/template-for-repository/proofs/Makefile-project-defines
+++ b/template-for-repository/proofs/Makefile-project-defines
@@ -13,10 +13,16 @@
 # Makefile.common.
 ################################################################
 
-# Name of this proof project, displayed in proof reports
+# Name of this proof project, displayed in proof reports. For example,
+# "s2n" or "Amazon FreeRTOS". For projects with multiple proof roots,
+# this may be overridden on the command-line to Make, for example
+#
+# 	  make PROJECT_NAME="FreeRTOS MQTT" report
+#
 # PROJECT_NAME =
 
-# Path to litani executable
+# Path to litani executable, relative to the root of the repository.
+# This applies even for projects with multiple proof roots.
 # LITANI =
 
 # Flags to pass to goto-cc for compilation (typically those passed to gcc -c)
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -244,33 +244,6 @@ SPACE :=$() $()
 COMMA :=,
 
 ################################################################
-# Useful macros for running commands
-
-#1: command, 2: flags, 3: log file
-DO_AND_LOG_COMMAND = $(1) $(2) 2>&1 | tee $(3); exit $${PIPESTATUS[0]}
-
-#1: command, 2: flags, 3: log file
-
-# CBMC uses the special error-code 10 to signify that it detected an
-# assertion violation.	Continue the build so we can output the trace.
-DO_AND_LOG_IGNORING_ERROR_10 =	$(1) $(2) 2>&1 | tee $(3); if [ $${PIPESTATUS[0]} -ne 10 ]; then exit $${PIPESTATUS[0]}; fi
-
-#1: flags, 2: source, 3: target
-DO_GOTO_ANALYZER = $(call DO_AND_LOG_COMMAND,$(GOTO_ANALYZER),$(CBMC_VERBOSITY) $(1) $(2) $(3), $(call LOG_FROM_ENTRY,$(3)))
-
-#1: flags, 2: source, 3: target
-DO_GOTO_CC =  $(call DO_AND_LOG_COMMAND,$(GOTO_CC),$(CBMC_VERBOSITY) $(1) $(2) -o $(3), $(call LOG_FROM_ENTRY,$(3)))
-
-#1: flags, 2: source, 3: target
-DO_GOTO_INSTRUMENT = $(call DO_AND_LOG_COMMAND,$(GOTO_INSTRUMENT),$(CBMC_VERBOSITY) $(1) $(2) $(3), $(call LOG_FROM_ENTRY,$(3)))
-
-#1: flags, 2: source, 3: logfile
-DO_CBMC =  $(call DO_AND_LOG_IGNORING_ERROR_10,$(CBMC),$(CBMC_VERBOSITY) --flush $(1) $(2), $(3))
-
-#1: message 2: source 3: dest
-DO_NOOP_COPY = cp $(2) $(3); echo $(1) | tee $(call LOG_FROM_ENTRY,$(3))
-
-################################################################
 # Useful macros translating filenames
 
 C_FROM_GOTO = $(patsubst $(GOTODIR)%,$(SRCDIR)%,$(1:.goto=.c))
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -300,51 +300,129 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
 
 # Compile project sources
 $(PROJECT_GOTO)1.goto: $(PROJECT_SOURCES)
-	mkdir -p $(dir $@)
-	mkdir -p $(dir $(call LOG_FROM_ENTRY,$@))
-	$(call DO_GOTO_CC,--export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES),$^,$@)
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_CC) $(CBMC_VERBOSITY) --export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES) $^ -o $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): building project binary"
 
 # Compile proof sources
 $(PROOF_GOTO)1.goto: $(PROOF_SOURCES)
-	mkdir -p $(dir $@)
-	mkdir -p $(dir $(call LOG_FROM_ENTRY,$@))
-	$(call DO_GOTO_CC,--export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES),$^,$@)
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_CC) $(CBMC_VERBOSITY) --export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES) $^ -o $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): building proof binary"
 
 # Optionally remove function bodies from project sources
 $(PROJECT_GOTO)2.goto: $(PROJECT_GOTO)1.goto
 ifeq ($(REMOVE_FUNCTION_BODY),"")
-	$(call DO_NOOP_COPY,"Not removing function bodies",$<,$@)
+	$(LITANI) add-job \
+	  --command 'cp $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@)\
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): not removing function bodies"
 else
-	$(call DO_GOTO_INSTRUMENT,$(CBMC_REMOVE_FUNCTION_BODY),$<,$@)
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) $(CBMC_REMOVE_FUNCTION_BODY) $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): removing function bodies"
 endif
 
-# Don't remove function bodies from proof sources
-$(PROOF_GOTO)2.goto: $(PROOF_GOTO)1.goto
-	$(call DO_NOOP_COPY,"Not removing function bodies",$<,$@)
-
 # Link project and proof sources into the proof harness
-$(HARNESS_GOTO)1.goto: $(PROOF_GOTO)2.goto $(PROJECT_GOTO)2.goto
-	$(call DO_GOTO_CC,--function $(HARNESS_ENTRY),$^,$@)
+$(HARNESS_GOTO)1.goto: $(PROOF_GOTO)1.goto $(PROJECT_GOTO)2.goto
+	$(LITANI) add-job \
+	  --command '$(GOTO_CC) --function $(HARNESS_ENTRY) $^ -o $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): linking project to proof"
 
 # Optionally fill static variable with unconstrained values
 $(HARNESS_GOTO)2.goto: $(HARNESS_GOTO)1.goto
 ifeq ($(NONDET_STATIC),"")
-	$(call DO_NOOP_COPY,"Not applying --nondet-static",$<,$@)
+	$(LITANI) add-job \
+	  --command 'cp $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@)\
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): not setting static variables to nondet"
 else
-	$(call DO_GOTO_INSTRUMENT,$(NONDET_STATIC),$<,$@)
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) $(NONDET_STATIC) $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): setting static variables to nondet"
 endif
 
 # Omit unused functions (sharpens coverage calculations)
 $(HARNESS_GOTO)3.goto: $(HARNESS_GOTO)2.goto
-	$(call DO_GOTO_INSTRUMENT,--drop-unused-functions,$<,$@)
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) --drop-unused-functions $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): dropping unused functions"
 
 # Omit initialization of unused global variables (reduces problem size)
 $(HARNESS_GOTO)4.goto: $(HARNESS_GOTO)3.goto
-	$(call DO_GOTO_INSTRUMENT,--slice-global-inits,$<,$@)
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) --slice-global-inits $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): omitting unused global initializations"
 
 # Final name for proof harness
 $(HARNESS_GOTO).goto: $(HARNESS_GOTO)4.goto
-	cp $< $@
+	$(LITANI) add-job \
+	  --command 'cp $< $@' \
+	  --inputs $^ \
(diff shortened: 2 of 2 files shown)


template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
-	$(call DO_GOTO_CC,--export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES),$^,$@)
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_CC) $(CBMC_VERBOSITY) --export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES) $^ -o $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
```
danielsn, line 297: This is more of a litani question, but I'm a bit surprised that outputs includes the log.

template-for-repository/proofs/Makefile.common
```diff
@@ -365,31 +421,105 @@ arpa: $(ARPA_BLDDIR)
...
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage test \
+	  --interleave-stdout-stderr \
+	  --stdout-file $@ \
+	  --ignore-returns 10 \
+	  --tags "stats-group:coverage computation" \
+	  --description "$(HARNESS_ENTRY): calculating coverage"
+
+_result: $(LOGDIR)/result.txt
+result:
```
danielsn, line 479: Why this pattern?

template-for-repository/proofs/Makefile.common
```diff
@@ -365,31 +421,105 @@ arpa: $(ARPA_BLDDIR)
...
+	$(MAKE) -B _property
+	$(LITANI) run-build
+
+_coverage: $(LOGDIR)/coverage.xml
+coverage:
+	$(LITANI) init --project $(PROJECT_NAME)
+	$(MAKE) -B _coverage
+	$(LITANI) run-build
+
+define VIEWER_CMD
```
danielsn, line 497: why?

template-for-repository/proofs/Makefile.common
```diff
@@ -399,17 +529,38 @@ $(LOGDIR)/source.json:
...
-	--coverage $(LOGDIR)/coverage.xml \
-	--property $(LOGDIR)/property.xml \
-	--srcdir $(SRCDIR) \
-	--goto $(HARNESS_GOTO).goto \
-	--reportdir report \
+_report2: $(LOGDIR)/result.xml $(LOGDIR)/property.xml $(LOGDIR)/coverage.xml
+	$(LITANI) add-job \
+	  --command "$$VIEWER2_CMD" \
+	  --inputs $^ \
+	  --outputs $(PROOFDIR)/report \
```
danielsn, line 550: Why is the logfile not an output here?

template-for-repository/proofs/Makefile.common
```diff
@@ -365,31 +421,105 @@ arpa: $(ARPA_BLDDIR)
...
+	    '$(CBMC) $(CBMC_VERBOSITY) --flush $(CBMCFLAGS) $(CHECKFLAGS) --unwinding-assertions --trace $<' \
+	  --inputs $^ \
+	  --outputs $@ \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage test \
+	  --interleave-stdout-stderr \
+	  --stdout-file $@ \
+	  --ignore-returns 10 \
+	  --tags "stats-group:safety checks" \
+	  --description "$(HARNESS_ENTRY): checking safety properties"
```
danielsn, line 435: Should we alphabatize the flags?

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): building project binary"
 
 # Compile proof sources
 $(PROOF_GOTO)1.goto: $(PROOF_SOURCES)
-	mkdir -p $(dir $@)
-	mkdir -p $(dir $(call LOG_FROM_ENTRY,$@))
-	$(call DO_GOTO_CC,--export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES),$^,$@)
+	$(LITANI) add-job \
```
danielsn, line 289: A lot of these calls have the same set of flags over and over. I wonder if there is a cleaner way to wrap that. Maybe a "standard_flags" set?

template-for-repository/proofs/Makefile.common
karkhaz: There are two reasons for this:
* Litani copies all output files to the report directory, so that they can be viewed in a web browser.
* Jobs that depend on an output of this job will not be invoked until this job terminates successfully. This doesn't matter in this particular case since nobody depends on this log. Further down, you'll notice that CBMC Viewer does depend on the logs of CBMC.

The build graph really is a graph, not an upside-down tree. Multiple outputs are fine, ninja will do the right thing.

template-for-repository/proofs/Makefile.common
karkhaz: If you like. I originally had them Christmas-tree sorted. It's up to you, I don't mind that much.
```
	$(LITANI) add-job \
	  --inputs $^ \
	  --outputs $@ \
	  --ci-stage test \
	  --stdout-file $@ \
	  --ignore-returns 10 \
	  --interleave-stdout-stderr \
	  --pipeline-name $(HARNESS_ENTRY) \
	  --tags "stats-group:safety checks" \
	  --description "$(HARNESS_ENTRY): checking safety properties" \
	  --command \
	    '$(CBMC) $(CBMC_VERBOSITY) --flush $(CBMCFLAGS) $(CHECKFLAGS) --unwinding-assertions --trace $<'
```
```python
#!/usr/bin/env python3

import sys

lines = sys.stdin.read().splitlines()
lines = sorted(lines)
lines = sorted(lines, key=lambda x: len(x))
print("\n".join(lines))
```

template-for-repository/proofs/Makefile.common
karkhaz: Request from @markrtuttle to be able to continue running `make report` or `make whatever` in individual proof directories and get the same experience. What this does is run Litani with just a single job, so you'll get the cute webpage with a single row on it, and you'll still get an error if the proof failed.

In the proof repositories, users will have a superior experience by using the proof running script. My s2n pull request allows you to do
```
./run-cbmc-proofs --proofs s2n_buffer_stuffer s2n_huffer_puffer
```
which will run just those two proofs, or run without any arguments to run all the proofs. But, developers can still choose to run `make report` in a proof directory if that's what they're used to.

template-for-repository/proofs/Makefile.common
karkhaz: I can take this out if you prefer. It's just so that the `_report` rule below doesn't get hideously long, I reference this variable on line 510.

template-for-repository/proofs/Makefile.common
karkhaz: You're right, I'll make it one. Thanks.

template-for-repository/proofs/Makefile.common
karkhaz: I thought about this, and in fact I originally used the macros that you had implemented (which are deleted above). In the end I erred on the side of 'explicit is better than implicit' and avoiding too much redirection, since Litani is an unfamiliar tool. Really the only two flags that are common across all jobs are `--inputs $^ --outputs $@ $(call LOG_FROM_ENTRY,$@)`, and there are a few more flags that are on most but not all jobs.

Up to you, I don't mind either reintroducing the macros or having variables, but this way read the most intuitively in my opinion.

template-for-repository/proofs/Makefile.common
karkhaz: TBH now that I've written a few of these, I'm starting to prefer not alphabetizing them. There's a meaningful order to the flags. `inputs` and `outputs` should go together. `pipeline-name` and `ci-stage` are mandatory, so it's nice for them to go at the top next to `inputs` and outputs`. This means that all of the non-mandatory flags are together at the bottom of the list, so you can easily scan through them and see what's special about each particular job. I wouldn't want something like `ignore-returns` to be close to the top just before `inputs`, because it's unusual and it needs to be noticed.

template-for-repository/proofs/Makefile.common
karkhaz: Done

template-for-repository/proofs/Makefile-project-defines
```diff
@@ -13,6 +13,12 @@
 # Makefile.common.
 ################################################################
 
+# Name of this proof project, displayed in proof reports
```
markrtuttle, line 16: What the name of this proof project?  Do you mean "Amazon FreeRTOS" or "s2n"?  I'd give an example.  

In the development branch of the aws/aws-iot-device-sdk-embedded-C, the starter kit it submoduled under tools, and the two projects libraries/standard/mqtt and libraries/standard/http have independent subdirectories cbmc/proofs with independent Makefile-project-defines.  Yet CI has no trouble finding all of these proofs and running them all together.  Are you expecting that there is a single top-level Makefile-project-defines for the whole source tree?

template-for-repository/proofs/Makefile-project-defines
```diff
@@ -13,6 +13,12 @@
...
+# PROJECT_NAME =
+
+# Path to litani executable
+# LITANI =
```
markrtuttle, line 20: Same question: Are you expecting this to live in a top-level file for the whole repository?

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
 
 # Compile project sources
 $(PROJECT_GOTO)1.goto: $(PROJECT_SOURCES)
-	mkdir -p $(dir $@)
-	mkdir -p $(dir $(call LOG_FROM_ENTRY,$@))
-	$(call DO_GOTO_CC,--export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES),$^,$@)
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_CC) $(CBMC_VERBOSITY) --export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES) $^ -o $@' \
```
markrtuttle, line 278: It was a mistake to hard code export-file-local-symbols in the original template.  No project MUST use it.  This should be part of the flags to CBMC that can be modified as needed.  Consider adding it above to COMPILE_FLAGS.

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
+	  --inputs $^ \
```
markrtuttle, line 279: It was a mistake to compile all inputs at once in the original template.  Doing that here may be too much to do in a single pull request.  But is it going to be a problem to fix this in the next pull request?

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
```
markrtuttle, line 284: I expected you would delete all of these macros.

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
 
 # Compile proof sources
 $(PROOF_GOTO)1.goto: $(PROOF_SOURCES)
-	mkdir -p $(dir $@)
-	mkdir -p $(dir $(call LOG_FROM_ENTRY,$@))
-	$(call DO_GOTO_CC,--export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES),$^,$@)
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_CC) $(CBMC_VERBOSITY) --export-file-local-symbols $(COMPILE_FLAGS) $(INCLUDES) $(DEFINES) $^ -o $@' \
+	  --inputs $^ \
```
markrtuttle, line 292: Comments about export-file-local-symbols and inputs above repeat here.

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
-	$(call DO_NOOP_COPY,"Not removing function bodies",$<,$@)
+	$(LITANI) add-job \
+	  --command 'cp $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@)\
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): not removing function bodies"
```
markrtuttle, line 311: not removing function bodies from project sources

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) $(CBMC_REMOVE_FUNCTION_BODY) $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): removing function bodies"
```
markrtuttle, line 322: removing function bodies from project sources.

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
 endif
 
-# Don't remove function bodies from proof sources
-$(PROOF_GOTO)2.goto: $(PROOF_GOTO)1.goto
-	$(call DO_NOOP_COPY,"Not removing function bodies",$<,$@)
```
markrtuttle, line 323: from proof sources

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
+$(HARNESS_GOTO)1.goto: $(PROOF_GOTO)1.goto $(PROJECT_GOTO)2.goto
+	$(LITANI) add-job \
+	  --command '$(GOTO_CC) --function $(HARNESS_ENTRY) $^ -o $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): linking project to proof"
```
markrtuttle, line 335: project and proof into harness

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
-# Don't remove function bodies from proof sources
-$(PROOF_GOTO)2.goto: $(PROOF_GOTO)1.goto
-	$(call DO_NOOP_COPY,"Not removing function bodies",$<,$@)
-
 # Link project and proof sources into the proof harness
-$(HARNESS_GOTO)1.goto: $(PROOF_GOTO)2.goto $(PROJECT_GOTO)2.goto
-	$(call DO_GOTO_CC,--function $(HARNESS_ENTRY),$^,$@)
+$(HARNESS_GOTO)1.goto: $(PROOF_GOTO)1.goto $(PROJECT_GOTO)2.goto
+	$(LITANI) add-job \
+	  --command '$(GOTO_CC) --function $(HARNESS_ENTRY) $^ -o $@' \
```
markrtuttle, line 328: verbosity?

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) --slice-global-inits $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@) \
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): omitting unused global initializations"
```
markrtuttle, line 386: Prefer phrasing here and elsewhere to reflect phrasing of the option.  As in "Slicing unused" or "Removing unused"

template-for-repository/proofs/Makefile.common
```diff
@@ -300,54 +273,137 @@ CBMC_REMOVE_FUNCTION_BODY := $(patsubst %,--remove-function-body %, $(REMOVE_FUN
...
+	  --inputs $^ \
+	  --outputs $@ $(call LOG_FROM_ENTRY,$@)\
+	  --pipeline-name $(HARNESS_ENTRY) \
+	  --ci-stage build \
+	  --interleave-stdout-stderr \
+	  --stdout-file $(call LOG_FROM_ENTRY,$@) \
+	  --description "$(HARNESS_ENTRY): copying final goto-binary"
+
+_goto: $(HARNESS_GOTO).goto
 	echo $(DEPENDENT_GOTOS)
```
markrtuttle, line 401: What is _goto and what are DEPENDENT_GOTOS?

template-for-repository/proofs/Makefile.common
```diff
@@ -365,31 +421,105 @@ arpa: $(ARPA_BLDDIR)
...
+	  --stdout-file $@ \
+	  --ignore-returns 10 \
+	  --tags "stats-group:safety checks" \
+	  --description "$(HARNESS_ENTRY): checking safety properties"
 
 $(LOGDIR)/property.xml: $(HARNESS_GOTO).goto
-	$(call DO_CBMC,$(CBMCFLAGS) $(CHECKFLAGS) --unwinding-assertions --show-properties --xml-ui,$<,$@)
+	$(LITANI) add-job \
+	  --command \
+	    '$(CBMC) $(CBMC_VERBOSITY) $(CBMCFLAGS) $(CHECKFLAGS) --show-properties --xml-ui $<' \
```
markrtuttle, line 452: Why omit --unwinding-assertions?  They are certainly properties we check.  Even if they don't actually show up in the output, and I think they do not, it is still important to visually pattern match the property checking invocation.

template-for-repository/proofs/Makefile.common
markrtuttle: Even to @markrtuttle this is a surprise.  A comment explaining "we want targets for the whole job AND individual steps" would be helpful.  Can you remove goto down here so that all of these @markrtuttle targets are defined in the same place?

template-for-repository/proofs/Makefile.common
```diff
@@ -365,31 +421,105 @@ arpa: $(ARPA_BLDDIR)
...
+	$(LITANI) run-build
+
+define VIEWER_CMD
+  $(VIEWER) \
+    --result $(LOGDIR)/result.txt \
+    --block $(LOGDIR)/coverage.xml \
+    --property $(LOGDIR)/property.xml \
+    --srcdir $(SRCDIR) \
+    --goto $(HARNESS_GOTO).goto \
+    --htmldir $(PROOFDIR)/html
```
markrtuttle, line 500: Different projects could want different combinations of arguments.  Eg, some may want to exclude parts of the source tree from annotation or prefer one symbol linking strategy over another.  At the very least we need a VIEWERFLAGS variable here.

template-for-repository/proofs/Makefile.common
```diff
@@ -365,31 +421,105 @@ arpa: $(ARPA_BLDDIR)
...
+	  --description "$(HARNESS_ENTRY): generating report"
+report:
+	$(LITANI) init --project $(PROJECT_NAME)
+	$(MAKE) -B _report
+	$(LITANI) run-build
+
+report:
+	$(LITANI) init --project $(PROJECT_NAME)
+	$(MAKE) -B _report
+	$(LITANI) run-build
```
markrtuttle, line 522: Two identical report targets.  I'd delete one.

template-for-repository/proofs/Makefile.common
```diff
@@ -399,17 +529,41 @@ $(LOGDIR)/source.json:
...
 
+define VIEWER2_CMD
+  $(VIEWER2) \
+    --result $(LOGDIR)/result.xml \
+    --coverage $(LOGDIR)/coverage.xml \
+    --property $(LOGDIR)/property.xml \
+    --srcdir $(SRCDIR) \
+    --goto $(HARNESS_GOTO).goto \
+    --reportdir $(PROOFDIR)/report
+endef
```
markrtuttle, line 540: Consider including a VIEWER2FLAGS variable here.

template-for-repository/proofs/Makefile.common
markrtuttle: All litani invocations in this file are hideously long.  :-). And as @danielsn pointed out elsewhere, there is enough duplication that the opportunity to change one place and not all places seems nonzero.

template-for-repository/proofs/Makefile-project-defines
karkhaz: Resolved during offline conversation, and added a clarifying comment.

template-for-repository/proofs/Makefile.common
karkhaz: I would like this PR to focus entirely on porting to Litani, without making any functional changes. Once we've deployed this to projects, we can experiment with changing the rules and check whether they have any effect on any of the proofs.

template-for-repository/proofs/Makefile.common
karkhaz: As above, I would like to restrict this PR to purely porting to Litani.

template-for-repository/proofs/Makefile.common
karkhaz: As above, I would like to restrict this PR to purely porting to Litani.

template-for-repository/proofs/Makefile.common
karkhaz: Done, thank you

template-for-repository/proofs/Makefile.common
karkhaz: Done

template-for-repository/proofs/Makefile.common
karkhaz: As above, I would like to restrict this PR to purely porting to Litani.

template-for-repository/proofs/Makefile.common
karkhaz: As above, I would like to restrict this PR to purely porting to Litani.

template-for-repository/proofs/Makefile.common
karkhaz: > there is enough duplication that the opportunity to change one place and not all places seems nonzero

not sure how there is duplication here? This is defining a single variable and using it once below, to reduce clutter. I do think it might be a good idea to have flags for viewer, but I believe that that should be a separate PR. Not sure what you'd like me to do here, keeping this for now.

template-for-repository/proofs/Makefile.common
karkhaz: Done, thank you

markrtuttle:
This merges into litani-mainline and not into master, is that a mistake?

markrtuttle:
The starter kit now depends on litani, is that right?   It is no longer an option NOT to use litani, right?  If so, then the starter kit needs to have litani as a submodule.

karkhaz:
> This merges into litani-mainline and not into master, is that a mistake?

Not a mistake. Details are in the original PR message above. I don't want to merge this into master until we switch off the old CI, so that projects can continue to use the master branch.

> The starter kit now depends on litani, is that right? It is no longer an option NOT to use litani, right? If so, then the starter kit needs to have litani as a submodule.

I don't think this is a good idea because projects will be using Litani for efforts other than CBMC, so may wish to put litani somewhere else in the source tree (outside of the templates directory). This is compounded in projects like [aws/aws-iot-device-sdk-embedded-C](https://github.com/aws/aws-iot-device-sdk-embedded-C) that have multiple proof roots, I'd rather not dictate where the litani directory lives. What would be better is to have a README explaining very clearly that litani is a dependency and where to find it.

PR #19, opened 2020-07-30
Add a validation script for Arpa

ArenBabikian:
*Description of changes:*
I have added a script that allows to evaluate and validate the effectiveness of `arpa` if it is being used as the sole source of dependencies for a given proof.

The [proposed validation script](https://github.com/ArenBabikian/aws-templates-for-cbmc-proofs/blob/master/template-for-repository/proofs/arpa-test.sh) compares the goto functions incorporated during proof runs with and without the use of `arpa`. After running each approach, I run the `cbmc --show-goto-functions gotos/harness_file_name.goto` command and compare the outputs (after post-processing). If the post-processed outputs are identical, then the script concludes that `arpa` was able to find all dependencies.

Currently, this validation script has been tested successfully on [S2N](https://github.com/awslabs/s2n).

### DIFF
--- a/template-for-repository/.gitignore
+++ b/template-for-repository/.gitignore
@@ -7,9 +7,11 @@ proofs/**/html
 # Emitted by CBMC Viewer
 TAGS-*
 
-# Emitted by Arpa
+# Emitted by Arpa and its validator
 arpa_cmake/
 Makefile.arpa
+arpa-test-results.log
+arpa-test-logs
 
 # These files should be overwritten whenever prepare.py runs
 cbmc-batch.yaml
--- a/template-for-repository/.gitignore
+++ b/template-for-repository/.gitignore
@@ -7,7 +7,7 @@ proofs/**/html
 # Emitted by CBMC Viewer
 TAGS-*
 
-# Emitted by Arpa and its validator
+# Emitted by Arpa and its validation script
 arpa_cmake/
 Makefile.arpa
 arpa-test-results.log
--- a/template-for-repository/proofs/arpa-test.sh
+++ b/template-for-repository/proofs/arpa-test.sh
@@ -3,29 +3,16 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-# This script performs validation and data collection for Arpa
-
-makefile=Makefile
-results=arpa-test-results.log
-csv=arpa-test-results.csv
-
-overwrite_files="s2n_free_object s2n_mem_cleanup s2n_pkcs3_to_dh_params s2n_stuffer_free"
-# test_files="s2n_stuffer_skip_to_char s2n_free_object s2n_stuffer_send_to_fd s2n_mem_cleanup s2n_pkcs3_to_dh_params s2n_stuffer_free s2n_stuffer_rewrite"
-
-arpa_log=arpa-test-logs
-makefile_std_save=$arpa_log/$makefile
-goto_functions_std=$arpa_log/goto-functions-std
-goto_functions_arpa=$arpa_log/goto-functions-arpa
-
-#data collection
-
-list_deps_arpa_cannot_find+="" 
-# list_deps_arpa_has_found_at_least_once+=""
-# list_deps_that_currently_exist_but_are_irrelevant+="" # case where arpa finds less deps but diff of show_goto_functions is the same
+# This script performs  data collection for Arpa.
+# Furthermore, it serves as a validation framework for existing CBMC proofs.
+manual_proof_override="s2n_free_object s2n_mem_cleanup s2n_pkcs3_to_dh_params s2n_stuffer_free"
 
 
 function initialize {
+    results=arpa-test-results.log
+    arpa_log=arpa-test-logs
     t_pfs_w_unnesc=0
+
     rm -f $results
     tot_prfs=$(ls -l | grep -c ^d)
 }
@@ -34,9 +21,10 @@ function initialize {
 function look_for {
     if [ ! -f $1 ]; then
         echo "-- $1 does not exist in ($dir)." 1>&2
-        echo "<SKIPPING>" 1>&2
+        echo "< SKIPPING >" 1>&2
         cd ..
-        echo "ERROR  -  $dir" >> $results
+        echo "($cnt/$tot_prfs) - ERROR - $dir" >> $results
+        echo -e "  0.- $1 does not exist in ($dir)\n" >> $results
         continue
     fi
 }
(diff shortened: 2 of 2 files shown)


template-for-repository/.gitignore
```diff
@@ -7,9 +7,11 @@ proofs/**/html
 # Emitted by CBMC Viewer
 TAGS-*
 
-# Emitted by Arpa
+# Emitted by Arpa and its validation script
 arpa_cmake/
 Makefile.arpa
+arpa-test-results.log
+arpa-test-logs
```
markrtuttle, line 14: Do you mean arpa-test-logs/

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
+#!/bin/bash
```
markrtuttle, line 1: Why a shell script and not a python script?

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# This script performs  data collection for Arpa.
+# Furthermore, it serves as a validation framework for existing CBMC proofs.
+
+# TODO add project-specific list of proofs that fail, but should be overriden due to limitations of the validation approach
```
markrtuttle, line 9: Talk to Kareem, but I'm not sure TODO lists belong in public repositories.  But I may violate this rule myself...

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+# The list below must be seperated by spaces
+manual_proof_override=""
```
markrtuttle, line 11: Can you say what this is?

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+
+
+function initialize {
+    # variable initializations
+    results=arpa-test-results.log
+    arpa_log=arpa-test-logs
+    t_pfs_w_unnesc=0
+    tot_prfs=$(ls -l | grep -c ^d)
```
markrtuttle, line 19: I prefer readable names like total_proofs.  

Is this supposed to be the total number of proofs in the proof repository?  This assumes all proofs are at the same level in the file system hierarchy.  This is incorrect in all the FreeRTOS work.

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+    fi
+}
+
+
+function look_for {
+    if [ ! -f $1 ]; then
+        echo "-- $1 does not exist in ($dir)." 1>&2
+        echo "< SKIPPING >" 1>&2
+        cd ..
+        echo "($cnt/$tot_prfs) - ERROR - $dir" >> $results
```
markrtuttle, line 38: No idea what is going on here, but this is more than just looking for a file.

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+        echo -e "  0.- $1 does not exist in ($dir)\n" >> $results
+        continue
+    fi
+}
+
+
+function proof_init {
+    # proof-specific initialisations
+    cd $dir
```
markrtuttle, line 47: This is why shell scripts are a problem.  Where is $dir defined?

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+    harness="$dir"_harness.c
```
markrtuttle, line 48: This is assuming a very particular naming scheme for the harness.  At the very least it needs to be weakened to just assuming that the harness ends in "_harness.c"

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+    # check if required files exist
+    look_for Makefile
+    look_for $harness
+}
+
+
+function get_fcts {
+    # this function gets the goto functions relevant to the CBMC proof
+    # and store it in a variable
+    goto_fcts=$(cbmc --show-goto-functions gotos/$harness.goto)
```
markrtuttle, line 61: I think $harness was defined above to end in .c, so the name of the goto binary is going to be *.goto and not *.c.goto.

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+
+    # get rid of comments
+    clean_goto_fcts=$(grep -v " *//.*" <<< "$goto_fcts")
```
markrtuttle, line 64: Wouldn't it be less brittle to get the json output and parse that with jq?

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+    eval fcts_$1_clean="\$clean_goto_fcts"
```
markrtuttle, line 65: You see to be invoking eval on unsanitized input $1.  That seems very dangerous.

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+function count_lines {
+    wc -l <<< "$1" | tr -d " "
+}
+
+
+function get_deps {
+    #this function defines variables for numeric evaluation at a later stage
+
+    # we consider all dependencies except the harness file itself
+    grep_out=$(grep 'PROOF_SOURCES += \$(PROOF_.*\|PROJECT_SOURCES += .*' $2)
```
markrtuttle, line 78: Why are we grepping the Makefile?  Isn't all of this stuff supposed to be found in the cmake output?

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
...
+    eval deps_$1="\$grep_out"
+    n_deps=$(count_lines "$grep_out")
```
markrtuttle, line 80: What is n_deps?  Is it number_of_dependents?

template-for-repository/proofs/arpa-test.sh
```diff
@@ -0,0 +1,342 @@
+#!/bin/bash
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# This script performs  data collection for Arpa.
+# Furthermore, it serves as a validation framework for existing CBMC proofs.
```
markrtuttle, line 7: What is the purpose of this script?  Is it to invoke Arpa?  Is it to validate Arpa?  Some of the script appears to be collecting data in order to compare the results of the Makefile with and without Arpa.

markrtuttle:
I'm not convinced that this test script belongs in the template repository.  Doesn't it belong in the Arpa repository itself that will itself be submoduled into proof repositories.  This test script looks like Arpa CI and doesn't feel like starter kit material.

ArenBabikian:
I agree with you that the Arpa validation belongs in the Arpa repository, and not in the template repository. @karkhaz agrees as well. Thus, I will close this PR and move the validation into the Arpa repo.

Thank you for your comments. I will assess them within the Arpa repo.

PR #22, opened 2020-08-31
Allow overriding expected CBMC result

SaswatPadhi:
This PR adds a new parameter in the Makefile that allows overriding `expected` (the expected result from CBMC) in the auto-generated `cbmc-batch.yaml` files.

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -446,9 +446,9 @@ cbmc-batch.yaml:
 	@echo 'build_memory: $(JOB_MEMORY)' > $@
 	@echo 'cbmcflags: $(strip $(call yaml_encode_options,$(CI_FLAGS)))' >> $@
 	@echo 'coverage_memory: $(JOB_MEMORY)' >> $@
-	@echo "expected: $(EXPECTED)" >> $@
-	@echo "goto: $(HARNESS_GOTO).goto" >> $@
-	@echo "jobos: $(JOB_OS)" >> $@
+	@echo 'expected: $(EXPECTED)' >> $@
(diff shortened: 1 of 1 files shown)


template-for-repository/proofs/Makefile.common
```diff
@@ -433,6 +433,7 @@ veryclean: clean
 
 JOB_OS ?= ubuntu16
 JOB_MEMORY ?= 32000
+EXPECTED ?= SUCCESSFUL
```
karkhaz, line 436: Please write a comment here. Suggest something like
```
Proofs that are expected to fail should set EXPECTED to
"FAILURE" in their Makefile. Values other than SUCCESSFUL
or FAILURE will cause a CI error.
```

karkhaz (changes requested):

template-for-repository/proofs/Makefile.common
SaswatPadhi: Thanks for the quick feedback. Done.

karkhaz: @danielsn @feliperodri I think this should actually be `FAILED` rather than `FAILURE`, right? My mistake.

karkhaz (changes requested):

template-for-repository/proofs/Makefile.common
feliperodri: Yep, CBMC reports VERIFICATION FAILED.

SaswatPadhi: Thanks, made then change.

karkhaz (approved):

PR #23, opened 2020-09-01
Negative tests for CBMC flags

SaswatPadhi:
This PR introduces a suite of negative tests. In future, we should see an alarm (a failed CI job) if any of these tests fail, which would indicate that CBMC was run with flags that were _too permissive_.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


SaswatPadhi:
I have addressed all the comments from the c-common PR and am talking to @eidelmanjonathan this afternoon regarding enabling CI on this repo.

template-for-repository/proofs/cbmc_ensure__bounds_check/Makefile
```diff
@@ -0,0 +1,17 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+
+UNWINDSET +=
+
+CBMCFLAGS +=
+
+EXPECTED = FAILED
+
+HARNESS_ENTRY = cbmc_ensure__bounds_check_harness
```
feliperodri, line 10: Why this pattern of two underscores in the harness' name (for all harnesses)?

template-for-repository/proofs/cbmc_ensure__signed_overflow_check/cbmc_ensure__signed_overflow_check_harness.c
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+/**
+ * A negative test for --signed-overflow-check flag
+ */
+void cbmc_ensure__signed_overflow_check_harness() {
+    int overflow, offset;
+    overflow += offset;
+    overflow -= offset;
```
feliperodri, line 12: We should split into two harnesses.

feliperodri (approved):
LGTM, only a few more small suggestions. The only questions is whether we all agree if that is the right place to keep these general negative harnesses.

markrtuttle (changes requested):
I do not think these negative tests belong in the starter kit.  They certainly don't belong in the top level of the proofs directory, they must be collected in a subdirectory of the proofs directory.  And the EXPECTED flag is unused by anything I'm aware of, so this is guaranteed to cause every pull request including these tests to fail CI.

SaswatPadhi:
@markrtuttle The `EXPECTED` flag actually works as expected. You can see a CI run here:https://github.com/awslabs/aws-c-common/pull/685.

I could move these to a subdirectory (`cbmc_check_flags`) within proofs, I think that's a good idea.

markrtuttle:
A name like "negative_tests" would be more accurate.

SaswatPadhi:
Thanks, I will go with that then.

template-for-repository/proofs/cbmc_ensure__bounds_check/Makefile
```diff
@@ -0,0 +1,17 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+
+UNWINDSET +=
```
markrtuttle, line 4: Delete in all Makefiles

template-for-repository/proofs/cbmc_ensure__bounds_check/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+CBMCFLAGS +=
```
markrtuttle, line 6: Delete in all Makefiles

template-for-repository/proofs/cbmc_ensure__bounds_check/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+EXPECTED = FAILED
+
+HARNESS_ENTRY = cbmc_ensure__bounds_check_harness
+HARNESS_FILE = $(HARNESS_ENTRY).c
+
+PROOF_SOURCES += $(HARNESS_FILE)
+
+PROJECT_SOURCES += $(HARNESS_FILE)
```
markrtuttle, line 15: Delete in all Makefiles.

markrtuttle (changes requested):
There must be some documentation telling the naive user what this is for.  To a beginner, these tests look just silly, because they are testing that by default all of the flags that are in the Makefile.common by default are present.  The issue is going to be when a user realizes that a default flag must be omitted, then what does the user do in this directory?  What do these tests add that a review of Makefile-project-defines does not already add?

template-for-repository/proofs/cbmc_ensure__bounds_check/Makefile
SaswatPadhi: I renamed the harnesses. Please check now.

template-for-repository/proofs/cbmc_ensure__signed_overflow_check/cbmc_ensure__signed_overflow_check_harness.c
SaswatPadhi: I made this change pushed several underflow harnesses (for other overflow flags as well), but now there is no 1:1 match between flags and negative tests, so disabling specific negative tests might be a little more complicated. Let's discuss this.

SaswatPadhi:
As per our last discussion, I have moved all the negative tests to a separate subdirectory (`negative_tests/`) that's directly under `template-for-repository/`.

I have also addressed all your comments, @feliperodri and @markrtuttle. Please take another look.

template-for-repository/negative_tests/float_overflow_check/float_overflow_check_harness.c
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+/**
+ * A negative test for --float-overflow-check flag
+ */
+void float_overflow_check_harness() {
+    float overflow, increment;
```
danielsn, line 10: Would there be value in also doing this for `double`? Or is it basically the same check?

template-for-repository/negative_tests/nan_check/nan_check_harness.c
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+/**
+ * A negative test for --nan-check flag
+ */
+void nan_check_harness() {
+    float nan;
+    nan = nan / nan;
```
danielsn, line 11: Could this also trigger the div by 0 error?

danielsn (approved):
Couple questions, but nothing that would block merging

markrtuttle (approved):

template-for-repository/negative_tests/nan_check/nan_check_harness.c
SaswatPadhi: Yes, this would trigger both. I wasn't sure how to _only_ trigger a `nan` (in this case it would be `0/0`).

template-for-repository/negative_tests/README.md
```diff
@@ -0,0 +1,13 @@
+## Negative Tests
+
+This directory contains negative checks to ensure that CBMC CI jobs are run
+with the complete set of property-checking flags
+(see `CHECKFLAGS` in [Makefile.common](../proofs/Makefile.common))
+which we consider to be part of the best practice.
+
+To enable these tests in CI jobs,
+copy the negative test directories into `../proofs`.
```
markrtuttle, line 9: copy the directory negative_tests to proofs/negative_tests.  (We don't want them to pollute the proofs directory with two dozen directories, just one.)

template-for-repository/negative_tests/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+If a property-checking flag is not used used by your project,
+you might want to disable the corresponding negative test.
+To do so, simply delete the particular negative test directory from `../proofs`.
```
markrtuttle, line 13: Delete the negative tests from proofs/negative_tests

markrtuttle (changes requested):

template-for-repository/negative_tests/README.md
SaswatPadhi: Good point! Thanks, lemme fix this.

markrtuttle (approved):

PR #26, opened 2020-09-04
Added new CBMC flags and support for disabling flags

SaswatPadhi:
*Description of changes:*
This PR introduces a new parameter called `DISABLE_FLAGS` that allows users to disable certain CBMC flags on their proofs. It also adds three pointer-related CBMC flags: `--malloc-fail-null`, `--malloc-may-fail`, and `--pointer-primitive-check`.

### DIFF
--- a/template-for-proof/Makefile
+++ b/template-for-proof/Makefile
@@ -1,13 +1,6 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-# Additional flags for CBMC
-CBMCFLAGS +=
-CHECKFLAGS +=
-
-# Disabled CBMC flags
-DISABLED_FLAGS =
-
 HARNESS_ENTRY = harness
 HARNESS_FILE = <__FUNCTION_NAME__>_harness
 
--- a/template-for-proof/Makefile
+++ b/template-for-proof/Makefile
@@ -1,8 +1,15 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-HARNESS_ENTRY=harness
-HARNESS_FILE=<__FUNCTION_NAME__>_harness
+# Additional flags for CBMC
+CBMCFLAGS +=
+CHECKFLAGS +=
+
+# Disabled CBMC flags
+DISABLED_FLAGS =
+
+HARNESS_ENTRY = harness
+HARNESS_FILE = <__FUNCTION_NAME__>_harness
 
 DEFINES +=
 INCLUDES +=
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -95,7 +95,7 @@ sinclude $(PROOF_ROOT)/Makefile-project-defines
 sinclude $(PROOF_ROOT)/Makefile-template-defines
 
 # SRCDIR is the path to the root of the source tree
-SRCDIR ?= $(abspath $(../..)
+SRCDIR ?= $(abspath ../..)
 
 # PROOFDIR is the path to the directory containing the proof harness
 PROOFDIR ?= $(abspath .)
@@ -110,6 +110,11 @@ CBMCDIR ?= $(PROOF_ROOT)/cbmc
 
 # Default CBMC flags used for property checking and coverage checking
 CBMCFLAGS += --unwind 1 $(CBMC_UNWINDSET)
+CBMCFLAGS += --malloc-may-fail
+CBMCFLAGS += --malloc-fail-null
+
+# Removed DISABLED_FLAGS from CBMCFLAGS
+CBMCFLAGS := $(filter-out $(DISABLED_FLAGS),$(CBMCFLAGS))
 
 # Additional CBMC flags used for property checking
 ifneq ($(strip $(EXTERNAL_SAT_SOLVER)),)
@@ -122,9 +127,13 @@ CHECKFLAGS += --float-overflow-check
 CHECKFLAGS += --nan-check
 CHECKFLAGS += --pointer-check
 CHECKFLAGS += --pointer-overflow-check
+CHECKFLAGS += --pointer-primitive-check
 CHECKFLAGS += --signed-overflow-check
 CHECKFLAGS += --undefined-shift-check
 CHECKFLAGS += --unsigned-overflow-check
+
+# Removed DISABLED_FLAGS from CHECKFLAGS
+CHECKFLAGS := $(filter-out $(DISABLED_FLAGS),$(CHECKFLAGS))
 
 # Additional CBMC flags used for coverage checking
 COVERFLAGS +=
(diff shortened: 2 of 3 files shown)


template-for-repository/proofs/Makefile.common
```diff
@@ -122,10 +127,14 @@ CHECKFLAGS += --float-overflow-check
 CHECKFLAGS += --nan-check
 CHECKFLAGS += --pointer-check
 CHECKFLAGS += --pointer-overflow-check
+CHECKFLAGS += --pointer-primitive-check
 CHECKFLAGS += --signed-overflow-check
 CHECKFLAGS += --undefined-shift-check
 CHECKFLAGS += --unsigned-overflow-check
 
+# Removed DISABLED_FLAGS from CHECKFLAGS
+CHECKFLAGS := $(filter-out $(DISABLED_FLAGS),$(CHECKFLAGS))
```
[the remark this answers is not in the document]
SaswatPadhi: Yes, I rebased (a copy of) my negative tests branch and checked that the proof goes through upon adding the tested flags to `DISABLED_FLAGS`.

template-for-repository/proofs/Makefile.common
```diff
@@ -110,6 +110,11 @@ CBMCDIR ?= $(PROOF_ROOT)/cbmc
 
 # Default CBMC flags used for property checking and coverage checking
 CBMCFLAGS += --unwind 1 $(CBMC_UNWINDSET)
+CBMCFLAGS += --malloc-may-fail
+CBMCFLAGS += --malloc-fail-null
```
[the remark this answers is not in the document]
SaswatPadhi: Good idea, just fixed this.

template-for-repository/proofs/Makefile.common
```diff
@@ -95,7 +95,7 @@ sinclude $(PROOF_ROOT)/Makefile-project-defines
 sinclude $(PROOF_ROOT)/Makefile-template-defines
 
 # SRCDIR is the path to the root of the source tree
-SRCDIR ?= $(abspath $(../..)
+SRCDIR ?= $(abspath ../..)
```
markrtuttle, line 98: Holy cow!  How did that ever work?!?

template-for-repository/proofs/Makefile.common
```diff
@@ -109,8 +109,13 @@ PROOFDIR ?= $(abspath .)
 CBMCDIR ?= $(PROOF_ROOT)/cbmc
 
 # Default CBMC flags used for property checking and coverage checking
+CBMCFLAGS += --malloc-fail-null
+CBMCFLAGS += --malloc-may-fail
 CBMCFLAGS += --unwind 1 $(CBMC_UNWINDSET)
 
+# Remove DISABLED_FLAGS from CBMCFLAGS
+CBMCFLAGS := $(filter-out $(DISABLED_FLAGS),$(CBMCFLAGS))
```
markrtuttle, line 117: I suspect we want something similar for CBMCFLAGS, CHECKFLAGS, and COVERFLAGS.

template-for-proof/Makefile
```diff
@@ -1,8 +1,15 @@
...
 # SPDX-License-Identifier: Apache-2.0
 
-HARNESS_ENTRY=harness
-HARNESS_FILE=<__FUNCTION_NAME__>_harness
+# Additional flags for CBMC
+CBMCFLAGS +=
+CHECKFLAGS +=
+
+# Disabled CBMC flags
+DISABLED_FLAGS =
```
markrtuttle, line 9: I don't think these three definitions (duplicates from the makefile.common) are needed here.  It might be prudent to add some documentation here explaining the use of DISABLED_FLAGS.  But if your code works, it works even when DISABLED_FLAGS is undefined.

template-for-repository/proofs/Makefile.common
markrtuttle, line 117: Your method, however, works only for flags without arguments and fails to work for flags with arguments.  Suppose I want to remove --unwind 1.

One approached three of us discussed off line was to define variables

MALLOC_CAN_FAIL=--malloc-fail-null
MALLOC_MAY_FAIL=--malloc-may-fail
POINTER_PRIMITIVE_CHECK=--pointer-primitive-check

CBMCFLAGS += $(MALLOC_CAN_FAIL)
CBMCFLAGS += $(MALLOC_MAY_FAIL)
CBMCFLAGS += $(POINTER_PRIMITIVE_CHECK)

This would allow an individual proof not wanting the --pointer-primitive-check to write a proof Makefile to include

undefine POINTER_PRIMITIVE_CHECK

See the [Undefine Directive](https://www.gnu.org/software/make/manual/html_node/Undefine-Directive.html).  

I recommend experimenting with this and rewriting Makefile.common in this style.  If it works, it would be a general solution doing a general service to the project.

markrtuttle (changes requested):

template-for-repository/proofs/Makefile.common
karkhaz: This won't work because the proof-specific Makefile is read before Makefile.common. At that point, there is no variable to undefine.

Suggestion: do this inside Makefile.common
```Makefile
MALLOC_CAN_FAIL ?= --malloc-fail-null
```

which will set `MALLOC_CAN_FAIL` iff it hasn't already been set. Folks who want to override it in a proof can do
```Makefile
MALLOC_CAN_FAIL := 
```
inside the proof Makefile. Furthermore, if people would like to disable a flag for the entire repo, they could do the same thing inside Makefile-project-defines.

So I suggest to put these `?=` definitions *after* including Makefile-project-defines, so that proof authors have a chance to disable them both per-proof and project-wide.

SaswatPadhi: Makes sense, thanks! I will make the change.

danielsn: One suggestion: use a standard naming scheme. `CBMC_FLAG_MALLOC_CAN_FAIL`

template-for-repository/proofs/Makefile.common
```diff
@@ -115,16 +115,34 @@ CBMCFLAGS += --unwind 1 $(CBMC_UNWINDSET)
...
-CHECKFLAGS += --conversion-check
-CHECKFLAGS += --div-by-zero-check
-CHECKFLAGS += --float-overflow-check
-CHECKFLAGS += --nan-check
-CHECKFLAGS += --pointer-check
-CHECKFLAGS += --pointer-overflow-check
-CHECKFLAGS += --signed-overflow-check
-CHECKFLAGS += --undefined-shift-check
-CHECKFLAGS += --unsigned-overflow-check
+
```
markrtuttle, line 118: I thought you and Kareem had a conversation saying that this could not work.

If this does work, there needs to be some documentation here telling the user that these are the flags we recommend, and how to omit flags that the proof project decides not to use.

markrtuttle (changes requested):

template-for-repository/proofs/Makefile.common
SaswatPadhi: So, the discussion was that having this following in `Makefile.common`

```
MALLOC_CAN_FAIL=--malloc-fail-null
```

and this in (harness-specific) `Makefile`:

```
undefine MALLOC_CAN_FAIL
```

won't work. This is a different solution, described by @karkhaz in a comment above.

Adding some documentation is a good idea. Should I add it here, or add an example within `template-for-proof`, or both?

SaswatPadhi: I added some description above these variables. Please take a look.

feliperodri (approved):
LGTM, including the documentation. We need to make sure the malloc flags work properly before update projects.

markrtuttle (approved):

PR #27, opened 2020-09-09
Use a separate variable for proof ID

karkhaz:
This commit introduces a new variable that proof-writers must fill in,
PROOF_ID. Previously, the unique identifier for the proof was equal to
the HARNESS_ENTRY variable, but this is inconsistent with how the
starter kit is meant to be used. After this commit, HARNESS_ENTRY shall
refer only to the name of the entry point of the harness, and will not
be used as a proof identifier.

**NOTE**: this is targeted at merging into `litani-mainline`, not `master`.

### DIFF
--- a/template-for-proof/Makefile
+++ b/template-for-proof/Makefile
@@ -4,6 +4,10 @@
 HARNESS_ENTRY=harness
 HARNESS_FILE=<__FUNCTION_NAME__>_harness
 
+# This should be a unique identifier for this proof, and will appear on the
+# Litani dashboard. It can be human-readable and contain spaces if you wish.
+PROOF_UID = <__FUNCTION_NAME__>
+
 DEFINES +=
 INCLUDES +=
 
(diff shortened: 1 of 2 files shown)


template-for-proof/Makefile
```diff
@@ -4,6 +4,10 @@
 HARNESS_ENTRY=harness
 HARNESS_FILE=<__FUNCTION_NAME__>_harness
 
+# This should be a unique identifier for this proof, and will appear on the
+# Litani dashboard. It can be human-readable and contain spaces if you wish.
+PROOF_ID = <...>
```
markrtuttle, line 9: This needs to have a default value, can we use <__FUNCTION_NAME__> for the default?

Can we use a name like PROOF_UID that emphasizes the need for a UID and not just an ID?

markrtuttle (changes requested):

template-for-proof/Makefile
karkhaz: Done

markrtuttle (approved):

PR #29, opened 2020-09-25
Performance debugging documentation for starter kit

natasha-jeppu:
*Issue #, if available:*

*Description of changes:*
This adds a markdown file detailing CBMC profiling and metric data collection used to reason about proof performance. It describes all metrics that are collected along with a few strategies for improving performance using metric data. It also provides a functional coverage use case for the metric data.


feliperodri:
@danielsn should we close this PR?

jimgrundy:
@danielsn, @feliperodri asked back in April if we should close this. I'm going to recommend we close it if neither your nor @markrtuttle object.

PR #32, opened 2020-10-09
Strip whitespace from command line input

markrtuttle:
This simple modification just uses strip() to strip whitespace from script input() from the command line.

### DIFF
--- a/scripts/setup-proof.py
+++ b/scripts/setup-proof.py
@@ -49,10 +49,10 @@ def main():
     logging.basicConfig(format='%(levelname)s: %(message)s')
 
     print("What is the function name?  ", end="")
-    function = input()
+    function = input().strip()
 
     print("What is the source file that defines the function?  ", end="")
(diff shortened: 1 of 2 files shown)


markrtuttle:
I'd be interested in hearing about a better way to do command line input(), something closer to readline or closer to bash shell path completion.

SaswatPadhi:
Having these two lines at the top of the Python file:

```
import readline

readline.parse_and_bind('tab: complete')
```

would enable <kbd>TAB</kbd> completion (for paths) on all existing `input()` calls without any other changes.

But the auto completion only works for paths at or below the current directory. So it won't do anything useful if you press <kbd>TAB</kbd> after typing `/` or `../` for example.

markrtuttle:
Closed in favor of https://github.com/awslabs/aws-templates-for-cbmc-proofs/pull/36.  I'll look into Saswat's readline advice.

PR #34, opened 2020-10-14
Add malloc flags to coverage checking.

markrtuttle:
We recently added the flags --malloc-may-fail and --malloc-fail-null to cbmc property checking by default.  The same flags need to be added to coverage checking, too.  

This patch adds the flags to coverage checking and does some minor cleanup of the Makefile.common.

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -115,7 +115,7 @@ ifneq ($(strip $(EXTERNAL_SAT_SOLVER)),)
    CHECKFLAGS += --external-sat-solver $(EXTERNAL_SAT_SOLVER)
 endif
 
-# Additional CBMC flags used for property checking
+# Property checking flags
 #
 # Each variable below controls a specific property checking flag
 # within CBMC. If desired, a property flag can be disabled within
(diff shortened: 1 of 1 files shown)


feliperodri (approved):
LGTM. I'd also emphasize that we could have an additional macro to only add these flags during coverage check, in case we want to remove/override them (e.g., when verification runtime is too slow); however, this would still lead to errors when reporting coverage, so I'd keep as it is.

template-for-repository/proofs/Makefile.common
```diff
@@ -125,9 +125,9 @@ endif
 #     CHECK_FLAG_POINTER_CHECK =
```
SaswatPadhi, line 125: I noticed that I made made a typo in my PR which introduced this comment. This line should be:

```
#     CHECK_FLAG_POINTER_CHECK :=
```

This would define the `CHECK_FLAG_POINTER_CHECK` immediately (which would be later detected by `?=`).

Sorry!

SaswatPadhi (approved):

template-for-repository/proofs/Makefile.common
SaswatPadhi: I just tested this on my machine. `=` also works :) So ignore this ^ comment.

danielsn (approved):

PR #28, opened 2020-09-16
Use Litani for running proofs

karkhaz:
This changes Makefile.common to invoke Litani when building proofs. Litani is a tool hosted at [awslabs/aws-build-accumulator](https://github.com/awslabs/aws-build-accumulator).

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #35, opened 2020-10-15
Add generic "run all proofs in parallel" script

karkhaz:
This commit adds a generic script that finds all proof directories and
adds a litani job for each of them, and then runs litani. This provides
a convenient way to run all CBMC proofs in parallel.

The script provides an option, `--no-standalone`, that allows the proofs
to be run as part of a larger set of litani jobs, for example running
all the proofs in multiple projects all at once.

This script should be usable as-is in many CBMC-using projects, without
further customization. Project owners need only add a value to the
PROJECT_NAME variable in Makefile-project-defines (this was already a
requirement even before this commit), so that the script can initialize
litani with the name of the project.

### DIFF
--- a/scripts/util.py
+++ b/scripts/util.py
@@ -21,6 +21,7 @@ COPY_INSTEAD = [
     "Makefile-project-targets",
     "Makefile-project-testing",
     ".gitignore",
+    "run-cbmc-proofs.py",
 ]
 
 ################################################################
--- a/template-for-repository/proofs/run-cbmc-proofs.py
+++ b/template-for-repository/proofs/run-cbmc-proofs.py
@@ -1,0 +1,258 @@
+#!/usr/bin/env python3
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+import argparse
+import asyncio
+import logging
+import math
+import os
+import pathlib
+import subprocess
+import sys
+
+
+DESCRIPTION = "Configure and run all CBMC proofs in parallel"
+
+# Keep the epilog hard-wrapped at 70 characters, as it gets printed
+# verbatim in the terminal. 70 characters stops here --------------> |
+EPILOG = """
+This tool automates the process of running `make report` in each of
+the CBMC proof directories. The tool calculates the dependency graph
+of all tasks needed to build, run, and report on all the proofs, and
+executes these tasks in parallel.
+
+The tool is roughly equivalent to doing this:
+
+        litani init --project "my-cool-project";
+
+        find . -name cbmc-proof.txt | while read -r proof; do
+            pushd $(dirname ${proof});
+
+            # The `make _report` rule adds a single proof to litani
+            # without running it
+            make _report;
+
+            popd;
+        done
+
+        litani run-build;
+
+except that it is much faster and provides some convenience options.
+The CBMC CI runs this script with no arguments to build and run all
+proofs in parallel. The value of "my-cool-project" is taken from the
+PROJECT_NAME variable in Makefile-project-defines.
+
+The --no-standalone argument omits the `litani init` and `litani
+run-build`; use it when you want to add additional proof jobs, not
+just the CBMC ones. In that case, you would run `litani init`
+yourself; then run `run-cbmc-proofs --no-standalone`; add any
+additional jobs that you want to execute with `litani add-job`; and
+finally run `litani run-build`.
+"""
+
+
+def get_project_name():
+    cmd = [
+        "make",
+        "-f", "Makefile.common",
+        "echo-project-name",
+    ]
+    logging.debug(" ".join(cmd))
+    proc = subprocess.run(cmd, universal_newlines=True, stdout=subprocess.PIPE)
+    if proc.returncode:
+        logging.warning("could not run make to determine project name")
+        sys.exit(1)
+    if not proc.stdout.strip():
+        logging.warning(
+            "project name has not been set; using generic name instead. "
+            "Set the PROJECT_NAME value in Makefile-project-defines to "
+            "remove this warning")
+        return "<PROJECT NAME HERE>"
+    return proc.stdout.strip()
+
+
+def get_args():
+    pars = argparse.ArgumentParser(
+        description=DESCRIPTION, epilog=EPILOG,
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+    for arg in [{
+            "flags": ["-j", "--parallel-jobs"],
+            "type": int,
+            "metavar": "N",
+            "help": "run at most N proof jobs in parallel",
+    }, {
+            "flags": ["--no-standalone"],
+            "action": "store_true",
+            "help": "only configure proofs: do not initialize nor run",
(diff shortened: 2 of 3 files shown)


template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -0,0 +1,268 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not
+# use this file except in compliance with the License. A copy of the License is
+# located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on
+# an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions
+# and limitations under the License.
```
markrtuttle, line 14: This is the long version of the license, we use the short form in all other files.

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -0,0 +1,268 @@
...
+
```
markrtuttle, line 15: This looks like the script from HTTP with the proof_marker modification, are there other changes to notice.

markrtuttle (changes requested):

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -0,0 +1,268 @@
...
+        logging.debug(" ".join(cmd))
+        proc = subprocess.run(cmd)
+        if proc.returncode:
+            logging.error("Failed to run litani init")
+            sys.exit(1)
+
+    proof_dirs = list(get_proof_dirs(
+        proof_root, args.proofs, args.proof_marker))
+    if not proof_dirs:
+        logging.error("No proof directories found")
```
SaswatPadhi, line 232: I think we should use `logging.critical` since we cannot recover from this error and we call `sys.exit(1)` next.

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -0,0 +1,268 @@
...
+    if ret is None or ret < 3:
+        return 1
+    return ret - 2
+
+
+def print_counter(counter):
+    print(
+        "\rConfiguring CBMC proofs: "
+        "{complete:{width}} / {total:{width}}".format(
+            **counter), end="", file=sys.stderr)
```
SaswatPadhi, line 147: Since we are printing to `sys.stderr`, why not use `logging.info()` here?

SaswatPadhi (changes requested):
I have some minor comments, mostly regarding logging.

scripts/util.py
```diff
@@ -21,6 +21,7 @@
     "Makefile-project-targets",
     "Makefile-project-testing",
     ".gitignore",
+    "run-cbmc-proofs.py",
```
SaswatPadhi, line 24: Are users expected to modify the Python script?
If so, I think we should add a brief comment regarding what's beyond the scope of the command-line flags and might require modifications within the script.

template-for-repository/proofs/run-cbmc-proofs.py
feliperodri:
```python
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -0,0 +1,268 @@
...
+    tasks = []
+    for _ in range(task_pool_size()):
+        task = asyncio.create_task(configure_proof_dirs(
+            proof_queue, counter))
+        tasks.append(task)
+
+    await proof_queue.join()
+
+    print_counter(counter)
+    print("", file=sys.stderr)
```
feliperodri, line 256: `logging.info()`?

template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: Done

template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: No other changes, but kissat is not included in this PR. As discussed previously, we'll leave that to a future PR.

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -0,0 +1,268 @@
...
+def get_project_name():
+    cmd = [
+        "make",
+        "-f", "Makefile.common",
+        "echo-project-name",
+    ]
+    logging.debug(" ".join(cmd))
+    proc = subprocess.run(cmd, universal_newlines=True, stdout=subprocess.PIPE)
+    if proc.returncode:
+        logging.warning("could not run make to determine project name")
```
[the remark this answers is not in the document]
karkhaz: logging.critical isn't really intended for userspace code. When running in privileged mode, logging.critical gets sent to ALL connected TTYs and is intended for messages that indicate a whole-system failure or crash. error is the right level for user programs that are about to abort or otherwise can't continue.

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -0,0 +1,268 @@
...
+            "metavar": "DIR",
+            "help": "only run proof in directory DIR (can pass more than one)",
+    }, {
+            "flags": ["--project-name"],
+            "metavar": "NAME",
+            "default": get_project_name(),
+            "help": "project name for report. Default: %(default)s",
+    }, {
+            "flags": ["--proof-marker"],
+            "metavar": "NAME",
```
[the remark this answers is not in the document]
karkhaz: Done

template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: logging.info ignores the `\r` character at the beginning and forcibly inserts a newline at the end, I don't want that. As I've written it here, this status line gets repeatedly overwritten rather than being printed on a new line every time.

template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: This is to flush the status line which has been printed to stderr with a `\r` character, so I'd rather just use print here.

karkhaz:
Addressed comments, but closing in favour of adding this to #28

template-for-repository/proofs/run-cbmc-proofs.py
SaswatPadhi: I see, thanks for the clarification.

template-for-repository/proofs/run-cbmc-proofs.py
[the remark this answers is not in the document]
SaswatPadhi: Is this behavior specific to our systems?

[The Python docs](https://docs.python.org/3/howto/logging.html) say this:

| ... | ... |
| - | - |
ERROR | Due to a more serious problem, the software has not been able to perform some function.
CRITICAL | A serious error, indicating that the program itself may be unable to continue running.

PR #38, opened 2020-10-19
Update setup scripts to work with litani.

markrtuttle:
This patch:

* Moves the definition of PROJECT_NAME to Makefile-template-defines along with SRCDIR and LITANI. The repository templates are copied (or symlinked) into place, but these are all project-level configuration variables that must be defined in order for Makefile.common to run.
* Adds documentation of these configuration variables to Makefile-template-defines.
* Fixes the definition of LITANI
* Cleans up the text of the questions asked by the setup scripts.
* Stops asking for the name of the proofs directory from the top-level setup script since the script itself creates the directory.

### DIFF
--- a/scripts/setup-proof.py
+++ b/scripts/setup-proof.py
@@ -22,7 +22,7 @@ def read_proof_template(filename):
 
 def write_proof_template(lines, filename, directory):
     with open(os.path.join(directory, filename), "w") as data:
-        data.write('\n'.join(lines)+'\n')
+        data.writelines(line + '\n' for line in lines)
 
 def rename_proof_harness(function, directory):
     shutil.move(os.path.join(directory, "FUNCTION_harness.c"),
--- a/scripts/setup-proof.py
+++ b/scripts/setup-proof.py
@@ -22,7 +22,7 @@ def read_proof_template(filename):
 
 def write_proof_template(lines, filename, directory):
     with open(os.path.join(directory, filename), "w") as data:
-        data.write('\n'.join(lines))
+        data.write('\n'.join(lines)+'\n')
 
 def rename_proof_harness(function, directory):
     shutil.move(os.path.join(directory, "FUNCTION_harness.c"),
--- a/scripts/setup.py
+++ b/scripts/setup.py
@@ -10,7 +10,30 @@ import os
 
 import util
 
-def create_makefile_template_defines(source_root, proof_root, litani):
+SRCDIR_TEXT = """
+# Absolute path to the root of the source tree.
+#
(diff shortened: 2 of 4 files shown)


scripts/setup-proof.py
```diff
@@ -22,7 +22,7 @@ def read_proof_template(filename):
 
 def write_proof_template(lines, filename, directory):
     with open(os.path.join(directory, filename), "w") as data:
-        data.write('\n'.join(lines))
+        data.write('\n'.join(lines)+'\n')
```
karkhaz, line 25: Why not just
```python
print('\n'.join(lines), file=data)
```
print appends a newline.

SaswatPadhi, line 25: Probably not a huge issue if we have a small number of lines, but this could be made more efficient. Instead of constructing a very large string using `join`, we could do:

```py
        data.writelines("{}\n".format(line) for line in lines)
```

This would just iterate over a generator object.

scripts/setup.py
```diff
@@ -10,33 +10,65 @@
 
 import util
 
-def create_makefile_template_defines(source_root, proof_root, litani):
+SRCDIR_TEXT = """
+# Absolute path to the root of the source tree.
```
karkhaz, line 14: 🍎 = It looks like you're actually setting this to a *relative* path below––relative to the proof directory

scripts/setup.py
```diff
@@ -10,33 +10,65 @@
...
+#
+SRCDIR ?= $(abspath $(PROOF_ROOT)/{})
+"""
+
+LITANI_TEXT = """
+# Absolute path to the litani script.
```
karkhaz, line 20: 🍎

karkhaz (changes requested):

scripts/setup.py
markrtuttle: The result after applying `$(abspath ...)` is an absolute path.  Computing the path from a path relative to PROOF_ROOT is what was already there.

scripts/setup.py
markrtuttle: Same.

scripts/setup.py
SaswatPadhi: Not sure, but I think Mark meant that the variable will ultimately store absolute path. And that's true because we call `abspath` below (within `make`).

SaswatPadhi (approved):

scripts/setup-proof.py
markrtuttle: Done.

markrtuttle: I followed Saswat's recommendation to use writelines().

karkhaz (approved):

PR #39, opened 2020-10-22
Make it possible to override the use of an external solver

markrtuttle:
The Makefile.common uses the external SAT solver given by the environment variable (or makefile variable) EXTERNAL_SAT_SOLVER for property checking.  This patch makes is possible to override the use of the external solver
and use the default solver for a particular proof (eg, in the event that minisat is actually faster than kissat).

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -121,9 +121,19 @@ CBMCDIR ?= $(PROOF_ROOT)/cbmc
 # Default CBMC flags used for property checking and coverage checking
 CBMCFLAGS += --unwind 1 $(CBMC_UNWINDSET) --flush
 
+# Do property checking with the external SAT solver given by
+# EXTERNAL_SAT_SOLVER.  Do coverage checking with the default solver,
+# since coverage checking requires the use of an incremental solver.
+#
+# For a particular proof, if the default solver is faster, do property
+# checking with the default solver by including this definition in the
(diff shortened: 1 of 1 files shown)


template-for-repository/proofs/Makefile.common
```diff
@@ -121,9 +121,19 @@ CBMCDIR ?= $(PROOF_ROOT)/cbmc
 # Default CBMC flags used for property checking and coverage checking
 CBMCFLAGS += --unwind 1 $(CBMC_UNWINDSET) --flush
 
+# Do property checking with the external SAT solver given by
+# EXTERNAL_SAT_SOLVER.  Do coverage checking with the default solver,
```
SaswatPadhi, line 125: Should we also document where this variable can be typically defined, something like:

```
Do property checking with the external SAT solver given by
EXTERNAL_SAT_SOLVER variable from the environment or Makefile-project-defines.
```

If this is not the best place to document it, please add it elsewhere, but a search for `EXTERNAL_SAT_SOLVER` on this repo currently returns only the `ifneq` line below and no documentation.

SaswatPadhi (approved):
Approved with a minor comment about documenting the `EXTERNAL_SAT_SOLVER` flag.

template-for-repository/proofs/Makefile.common
markrtuttle: Done.  Thanks!

PR #41, opened 2020-11-09
Make run script return zero on proof failure

karkhaz:
Prior to this commit, litani would return a non-zero code when any proof
job failed, and the run script would throw an exception when this
happened.

This commit adds the --exit-zero-on-fail flag to the litani run-build
invocation, so that litani returns zero even if there are proof failures
(but will continue to return a non-zero exit code if there was a crash
or some other error).

This is so that CI can distinguish between proof failures (which can be
detected in the litani JSON output), and litani crashes due to bugs in
litani, inconsistent builds, or other errors.

### DIFF
--- a/template-for-repository/proofs/run-cbmc-proofs.py
+++ b/template-for-repository/proofs/run-cbmc-proofs.py
@@ -160,7 +160,7 @@ def get_proof_dirs(proof_root, proof_list, proof_marker):
 
 
 def run_build(litani, jobs):
-    cmd = [str(litani), "run-build"]
+    cmd = [str(litani), "run-build", "--exit-zero-on-fail"]
     if jobs:
         cmd.extend(["-j", str(jobs)])
 


SaswatPadhi (approved):

tegansb (approved):

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -160,7 +160,7 @@ def get_proof_dirs(proof_root, proof_list, proof_marker):
 
 
 def run_build(litani, jobs):
-    cmd = [str(litani), "run-build"]
+    cmd = [str(litani), "run-build", "--exit-zero-on-fail"]
```
markrtuttle, line 163: Does this mean we have to update the run script (which is copied and not linked) in every repository based on litani (including repositories based on the starter kit)?  Would it not be cleaner to make this option the default for litani?

markrtuttle (changes requested):

template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: > Does this mean we have to update the run script (which is copied and not linked) in every repository based on litani

Yes. Note that if I updated litani as you suggest below, we'd still have to post a PR bumping the submodule pointer on every repo.

> Would it not be cleaner to make this option the default for litani?

I'd rather not. I don't think returning zero on failure is the expected intuitive behavior.

SaswatPadhi: +1 on not returning zero on failures (by default).

Although, instead of adding this flag to `cmd` why can't we simply remove the `proc.returncode` check at line 169?

karkhaz: After an offline discussion, we've decided that litani should actually return zero on pipeline failure. Litani users will have to bump their submodule; I'll take care of PRs for that. Closing this PR.

karkhaz: (But to answer your question: proof failures are distinct from litani crashing or otherwise aborting, so we still want to catch the latter case even if proof failures return zero)

SaswatPadhi: Thanks, that makes sense.

PR #43, opened 2020-11-23
Allow proof to be run twice in multiple configs

karkhaz:
This commit introduces a new variable that can be set in the environment
by the user, PROOF_SUFFIX. The value of variable gets appended to the
name of every proof. This allows the same proofs to be run multiple
times in different configurations in the same litani run. See the
documentation for an example.


markrtuttle (changes requested):
Let's not modify the starter kit just for an experiment.  Let's spin up an EC2 instance, and run the run-cbmc-proofs.py script twice, once with EXTERNAL_SAT_SOLVER set and once with it unset.  The coreMQTT is a good repo to run the experiment on, I think.

PR #46, opened 2021-02-10
Add coverage-checking functionality

karkhaz:
This PR:
* Adds a `check-coverage` program that checks CBMC's coverage for a particular proof against an expected coverage value
* Allows users to define the expected coverage value for a proof in the proof's Makefile
* Adds a Litani job that does the check

**The functionality in this PR depends on version 2 of CBMC viewer.**

Users can now define the `EXPECTED_COVERAGE` variable in proof Makefiles.  If this variable is defined, Litani will run the check-coverage program on the value of `EXPECTED_COVERAGE` to determine whether the proof's coverage was as high as expected.

If the `EXPECTED_COVERAGE` variable is _not_ defined for a proof, the coverage check is skipped. However, if the `ENFORCE_COVERAGE_CHECKS` variable is defined for the whole repository, then proofs that have not defined `EXPECTED_COVERAGE` will instead fail.

### DIFF
--- a/template-for-repository/proofs/check-coverage
+++ b/template-for-repository/proofs/check-coverage
@@ -1,0 +1,84 @@
+#!/usr/bin/env python3
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+import argparse
+import decimal
+import json
+import pathlib
+import sys
+
+
+DESCRIPTION = """
+Return 0 if CBMC coverage for a proof is as high as expected, 1 otherwise
+"""
+
+EPILOG = """
+This script checks reads the JSON coverage file emitted when cbmc-viewer-2 runs
+on CBMC's output files. It compares the coverage value in the JSON file with the
+user-specified expected coverage number.  If CBMC's coverage is below the
+expected value, this script exits abnormally.
+
+CBMC's coverage number is used to ensure that CBMC has explored the code as
+completely as possible, and that the assumptions made in the proof are not so
+restrictive that they prevent CBMC from exercising all code paths. A drop in
+coverage compared to the expected result typically calls for proof-writers to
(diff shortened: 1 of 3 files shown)


template-for-repository/proofs/check-coverage
```diff
@@ -0,0 +1,84 @@
...
+
+    # The value in the cbmc-viewer JSON file is a float (0.0--1.0). Convert it
+    # into a percentage without incorrectly truncating the fractional part
+    ctx = decimal.Context(prec=2, rounding=decimal.ROUND_HALF_EVEN)
+    cov_dec = ctx.create_decimal(actual_cov)
+    cov_percent = cov_dec * 100
+
+    print("Expected %d%%, got %d%%" % (args.expected_coverage, cov_percent))
+
+    if cov_percent != args.expected_coverage:
```
mww-aws, line 79: Would this throw an error any time the coverage changes, not just any time it is below the threshold?  This seems like it will be a maintenance issue every time someone changes the code and the coverage changes slightly.  Would it be better to specify a minimum threshold and use that to determine whether to alarm?

Also, it does not seem to match the documentation, which would expect a <= rather than a !=

mww-aws (changes requested):
Generally looks good.  I think there is a typo at line 79 in check-coverage.

feliperodri (approved):
LGTM modulo @mww-aws request.

template-for-repository/proofs/check-coverage
feliperodri: @karkhaz any updates on this comment?

karkhaz: This is fine, but we're not intending to merge this PR, part of is is expected to go into CBMC viewer.

karkhaz:
I'm going to close this as there are no immediate plans to merge it. Once we're ready I can post something that we actually intend to merge.

PR #47, opened 2021-02-18
Add timeout for CBMC jobs

karkhaz:
This is to prevent the proof run from appearing to "hang" in CI when
users submit proofs that take an abnormally long time. As of this
commit, Litani will forcefully abort long-running CBMC jobs, and the
proof will display as having failed on the dashboard. The report for
that proof will also not be generated, since CBMC will not have
completed printing its output.


feliperodri (approved):
LGTM.

PR #50, opened 2021-05-17
Add check for proof UID uniqueness

karkhaz:
The run script now ensures that every proof has a unique, non-empty UID.
This is to prevent the jobs for two different proofs being added to the
same Litani pipeline, possibly causing a race between two proofs or
making the results of a passing proof mask the results of a failing one.


feliperodri (approved):
LGTM

SaswatPadhi (approved):
Nice! LGTM.

PR #51, opened 2021-05-19
Write Litani output files to local directory

karkhaz:
When using versions of Litani >=1.7.0, the run script will now make
Litani write its output files to a directory underneath the proof
directory rather than under $TMPDIR. In addition, the latest proof run
will always be symlinked from $PROOFDIR/output/latest, which means that
users can keep $PROOFDIR/output/latest/html/index.html open in their
browser window and reload the page whenever re-running the run script.

### DIFF
--- a/template-for-repository/.gitignore
+++ b/template-for-repository/.gitignore
@@ -3,6 +3,7 @@ proofs/**/logs
 proofs/**/gotos
 proofs/**/report
 proofs/**/html
+proofs/output
 
 # Emitted by CBMC Viewer
 TAGS-*
--- a/template-for-repository/proofs/run-cbmc-proofs.py
+++ b/template-for-repository/proofs/run-cbmc-proofs.py
@@ -60,6 +60,7 @@ HTML dashboard from the latest Litani run will always be symlinked to
 `output/latest/html/index.html`, so you can keep that page open in
 your browser and reload the page whenever you re-run this script.
 """
+# 70 characters stops here ----------------------------------------> |
 
 
 def get_project_name():
--- a/template-for-repository/proofs/run-cbmc-proofs.py
+++ b/template-for-repository/proofs/run-cbmc-proofs.py
@@ -278,6 +278,19 @@ async def main():
 
     if not args.no_standalone:
         cmd = [str(litani), "init", *init_pools, "--project", args.project_name]
+
+        if "output_directory_flags" in litani_caps:
+            out_prefix = proof_root / "output"
+            out_symlink = out_prefix / "latest"
+            out_index = out_symlink / "html" / "index.html"
+            cmd.extend([
+                "--output-prefix", str(out_prefix),
+                "--output-symlink", str(out_symlink),
+            ])
+            print(
+                "For your convenience, the output of this run will be "
+                "symbolically linked to %s" % str(out_index))
+
         logging.debug(" ".join(cmd))
         proc = subprocess.run(cmd)
         if proc.returncode:
--- a/template-for-repository/proofs/run-cbmc-proofs.py
+++ b/template-for-repository/proofs/run-cbmc-proofs.py
@@ -53,6 +53,12 @@ just the CBMC ones. In that case, you would run `litani init`
 yourself; then run `run-cbmc-proofs --no-standalone`; add any
 additional jobs that you want to execute with `litani add-job`; and
 finally run `litani run-build`.
+
+The litani dashboard will be written under the `output` directory; the
+cbmc-viewer reports remain in the `$PROOF_DIR/report` directory. The
+HTML dashboard from the latest Litani run will always be symlinked to
+`output/latest/html/index.html`, so you can keep that page open in
+your browser and reload the page whenever you re-run this script.
 """
 
 
@@ -288,8 +294,8 @@ async def main():
                 "--output-symlink", str(out_symlink),
             ])
             print(
-                "For your convenience, the output of this run will be "
-                "symbolically linked to %s" % str(out_index))
+                "\nFor your convenience, the output of this run will be "
+                "symbolically linked to %s\n" % str(out_index))
 
         logging.debug(" ".join(cmd))
         proc = subprocess.run(cmd)


template-for-repository/.gitignore
```diff
@@ -3,6 +3,7 @@ proofs/**/logs
 proofs/**/gotos
 proofs/**/report
 proofs/**/html
+proofs/output
```
markrtuttle, line 6: What is the difference between the output directory and the report directories?  Will the report for an individual proof for 'make report' go into the report directory?  Will the report for run-cbmc-proofs.py go into the output directory?  This feature seemed to be geared toward CI (or the person simulating CI) and not to the day-to-day developer of the proof.  Is it the case that all this pull request is doing is creating a symbolic link from the proofs directory to the /tmp directory containing the litani report?  I don't have a good experience with using the litani report while the script is running, because the links change so rapidly that the files are gone by the time I try to load them.

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -234,8 +247,23 @@ async def main():
...
+            out_symlink = out_prefix / "latest"
+            out_index = out_symlink / "html" / "index.html"
+            cmd.extend([
+                "--output-prefix", str(out_prefix),
+                "--output-symlink", str(out_symlink),
+            ])
+            print(
+                "For your convenience, the output of this run will be "
+                "symbolically linked to %s" % str(out_index))
+
```
markrtuttle, line 266: The problem with notes like this is that they are lost in the blizzard of output generated by run-cbmc-proofs.py

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -234,8 +247,23 @@ async def main():
     proof_root = pathlib.Path(__file__).resolve().parent
     litani = get_litani_path(proof_root)
 
+    litani_caps = get_litani_capabilities(litani)
```
markrtuttle, line 250: Is this capabilities file something that is shipped with litani or is it something the start kit user is supposed to construct?

karkhaz: It's not a file, this method is running `litani print-capabilities`, which makes Litani print out a JSON list of the features that it supports. This is because we don't know which version of Litani is going to be used with the starter kit, so the starter kit has to test whether Litani has certain features before using them.
```
> ~/code/litani/litani print-capabilities
[
  "atomic_report_update",
  "report_expire",
  "dir_lock_api",
  "dir_lock_api_v2",
  "outcome_table",
  "output_directory_flags",
  "pools"
]
```

template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: Not much I can do about this here, unless you have suggestions, but I've also documented this in the prolog, so people can read it if they run `-h`. See "Document the presence of the `output` directory"

template-for-repository/.gitignore
karkhaz: See "Document the presence of the `output` directory", hopefully that clarifies some of this.

> What is the difference between the output directory and the report directories?

Report directories are where CBMC-Viewer writes its output. They are opaque to Litani.
`output` is where Litani writes all of its files. Each run of Litani will be a subdirectory of `output`, and in addition, the directory `output/latest` will be a symbolic link to the most recent such subdirectory. So:
```
\_output
| |
| \_ 2021-05-21T10:00:00
| | |_ html
| | |_ run.json
| | \_ [other irrelevant litani stuff]
| |
| \_ 2021-05-21T10:30:00
| | |_ html
| | \_ run.json
| |
| \_ 2021-05-21T10:40:00
| | |_ html
| | \_ run.json
| |
| \_ latest --> 2021-05-21T10:40:00
```

karkhaz: I wrote a new comment in the epilog below:
```
The litani reports will be located under the `output` directory. The
HTML report from the latest Litani run will always be symlinked to
`output/latest/html/index.html`, so you can keep this page open in
your browser and reload the page whenever you re-run this script.
```

markrtuttle: The command `git grep 'Document the presence'  .` yields nothing, I don't know what you want me to look at.  Would you mind changing this sentence to "The litani reports will be located under the `output` directory (the cbmc-viewer reports will remain in the `reports` directory)."  The problem is that the word "report" is used all over, and because "make report" has traditionally meant make the cbmc-viewer report, the word report is becoming ambiguous.

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -52,6 +53,11 @@
 yourself; then run `run-cbmc-proofs --no-standalone`; add any
 additional jobs that you want to execute with `litani add-job`; and
 finally run `litani run-build`.
+
+The litani reports will be located under the `output` directory. The
```
markrtuttle, line 57: Would you mind changing this sentence to "The litani reports will be located under the output directory (the cbmc-viewer reports will remain in the reports directory)." The problem is that the word "report" is used all over, and because "make report" has traditionally meant make the cbmc-viewer report, the word report is becoming ambiguous.

template-for-repository/proofs/run-cbmc-proofs.py
markrtuttle: But some output is crucial (where to find the report and where to find the current progress) and most output is noise.  At the very least, surrounding the crucial output by blank lines would make it stand out.

template-for-repository/.gitignore
markrtuttle: Ignore this "git grep" comment, you were naming the commit to study.  Thanks.

template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: Changed text below. I think using "dashboard" for litani and "report" for cbmc-viewer is probably helpful, I'll start using that terminology from now on.
> The litani dashboard will be written under the `output` directory; the
cbmc-viewer reports remain in the `$PROOF_DIR/report` directory. The
HTML dashboard from the latest Litani run will always be symlinked to
`output/latest/html/index.html`, so you can keep that page open in
your browser and reload the page whenever you re-run this script.

template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: Done

markrtuttle (approved):

PR #54, opened 2021-06-16
Added new variable to enable using loop contracts

SaswatPadhi:
*Description of changes:*

Added a new flag `APPLY_LOOP_CONTRACTS` to be able to enforce loop contracts on functions that do not have function contracts.

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -305,22 +305,28 @@ PROOF_SOURCES ?=
 # report, making the proof run appear to "hang".
 CBMC_TIMEOUT ?= 21600
 
-# Proof writers could add function contracts or loop invariants in
-# CBMC proofs. They should follow three distinct processes:
-# 1. Check whether a function contract is sound, where CHECK_FUNCTION_CONTRACTS
-#    contains the list of functions with contracts to be checked.
-# 2. Replace function calls with their correspondent function contracts,
-#    where USE_FUNCTION_CONTRACTS contains the list of functions with
-#    contracts to be replaced. One must check separately whether a function
-#    contract is sound before replacing it.
-# 3. Apply loop invariants during verification, where APPLY_LOOP_INVARIANTS
-#    contains the list of functions with loop invariants to be applied.
-#    This option is currently unavailable.
+# Proof writers could add function contracts in their source code.
+# These contracts are ignored by default, but may be enabled in two distinct
+# contexts using the following two variables:
+# 1. To check whether one or more function contracts are sound with respect to
+#    the function implementation, CHECK_FUNCTION_CONTRACTS should be a list of
+#    function names.
+# 2. To replace calls to certain functions with their correspondent function
+#    contracts, USE_FUNCTION_CONTRACTS should be a list of function names.
+#    One must check separately whether a function contract is sound before
+#    replacing it in calling contexts.
 CHECK_FUNCTION_CONTRACTS ?= ""
 CBMC_CHECK_FUNCTION_CONTRACTS := $(patsubst %,--enforce-contract %, $(CHECK_FUNCTION_CONTRACTS))
 
 USE_FUNCTION_CONTRACTS ?= ""
 CBMC_USE_FUNCTION_CONTRACTS := $(patsubst %,--replace-call-with-contract %, $(USE_FUNCTION_CONTRACTS))
+
+# Similarly, proof writers could also add loop contracts in their source code
+# to obtain unbounded correctness proofs. Unlike function contracts, loop
+# contracts are not reusable and thus are checked and used simultaneously.
+# These contracts are also ignored by default, but may be enabled by setting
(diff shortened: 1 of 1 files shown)


feliperodri (approved):
LGTM.

template-for-repository/proofs/Makefile.common
```diff
@@ -305,23 +305,26 @@ PROOF_SOURCES ?=
...
+#    replacing it in calling contexts.
+# 3. To use loop contracts (invariants and variants) for unbounded verification,
+#    USE_LOOP_CONTRACTS should be set to "TRUE".
 CHECK_FUNCTION_CONTRACTS ?= ""
 CBMC_CHECK_FUNCTION_CONTRACTS := $(patsubst %,--enforce-contract %, $(CHECK_FUNCTION_CONTRACTS))
 
 USE_FUNCTION_CONTRACTS ?= ""
 CBMC_USE_FUNCTION_CONTRACTS := $(patsubst %,--replace-call-with-contract %, $(USE_FUNCTION_CONTRACTS))
 
+USE_LOOP_CONTRACTS ?= FALSE
```
feliperodri, line 326: Let's apply the new macro from https://github.com/diffblue/cbmc/pull/6184. We can only update starter kits in projects once CBMC's PR get merge and this will only break builds that tries to use it without support.

feliperodri (changes requested):

template-for-repository/proofs/Makefile.common
SaswatPadhi: I don't think I understand this comment about macro. What do CBMC macros have to do with Makefiles? ...

SaswatPadhi: > We can only update starter kits in projects once CBMC's PR get merge and this will only break builds that tries to use it without support.

But builds shouldn't be affected if projects don't push the starter kit pointer forward. Function contract support was added in a previous commit and most repos have already forwarded the pointer.

I wanted to merge this early so that the interns wouldn't have to maintain a separate starter-kit branch _and_ a separate project branch and so on. What do you suggest?

template-for-repository/proofs/Makefile.common
```diff
@@ -305,23 +305,26 @@ PROOF_SOURCES ?=
...
+# distinct flags:
+# 1. To check whether one or more function contracts are sound with respect to
+#    the function implementation, CHECK_FUNCTION_CONTRACTS should be a list of
+#    function names.
+# 2. To replace calls to certain functions with their correspondent function
+#    contracts, USE_FUNCTION_CONTRACTS should be a list of function names.
+#    One must check separately whether a function contract is sound before
+#    replacing it in calling contexts.
+# 3. To use loop contracts (invariants and variants) for unbounded verification,
+#    USE_LOOP_CONTRACTS should be set to "TRUE".
```
markrtuttle, line 319: This text could be very confusing.  Why is there a check and use function contract and no check and use loop contracts.  (I'm going to have trouble getting used to saying loop contract in place of loop invariant, do we really want to do this?)

template-for-repository/proofs/Makefile.common
markrtuttle, line 326: Why are we setting use loop contracts to false by default and not undefining it by default as we do use and check function contracts?  Turning on functions checking requires just setting the variable, but turning on loop checking requires setting the variable to TRUE.   The implementations should be identical.

markrtuttle (changes requested):

template-for-repository/proofs/Makefile.common
SaswatPadhi: 1. Because check+replace is rolled together for loops. We do check initialization and inductiveness for any provided invariant. It's not useful to separate the check+replace for loop contracts because, unlike function contracts, they cannot be reused.
2. Loop invariants are not the only loop contracts. We already have s2n use cases where we at least need assigns clause. Then we have loop variants coming soon for termination proofs. Those are all enabled by the same flag, because they are interlinked.

template-for-repository/proofs/Makefile.common
SaswatPadhi: By undefining do you mean:

```make
USE_LOOP_CONTRACTS ?= 
```

I agree, this could work too.

template-for-repository/proofs/Makefile.common
SaswatPadhi: I split this paragraph and now I talk about function contracts and loop contracts separately.

template-for-repository/proofs/Makefile.common
SaswatPadhi: Assuming that's what you meant, I modified the initialization.

markrtuttle (approved):

feliperodri (approved):
LGTM.

template-for-repository/proofs/Makefile.common
SaswatPadhi: Changed this to `1` in place of `TRUE`.

feliperodri (approved):

PR #56, opened 2021-07-07
Don't display runtime graphs for report generation

karkhaz:
This is because, when using CBMC Viewer 2, all reports are generated
within seconds; there's no longer any value in rendering the graph.


feliperodri (approved):
LGTM.

PR #57, opened 2021-07-09
Change run script --proof-marker to --marker-file

karkhaz:
This is because --proofs and --proof-marker begin with the same letters,
and furthermore users often type --proof instead of --proof. By changing
--proof-marker to --marker-file, the Python argparser will autocomplete
--proof to --proofs for user convenience.


SaswatPadhi (approved):
Thanks; LGTM

PR #67, opened 2021-09-08
Make Makefile check for CBMC Viewer version

karkhaz:
This commit makes the run script query CBMC Viewer's version to decide
which target to run `make` with. This is so that the script can be used
in environments that have either Viewer 1 or Viewer 2 installed without
any other user intervention.

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -797,18 +797,15 @@ coverage:
 	$(MAKE) -B _coverage
 	$(LITANI) run-build
 
-_report: $(PROOFDIR)/html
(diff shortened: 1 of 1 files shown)


markrtuttle (commented):
Could this be done in the Makefile itself?  Could "make report" run either "make report1" or "make report2" depending on whether 'cbmc-viewer --version' succeeds?

karkhaz:
Done

markrtuttle:
I don't think this works as you expect.  The run script calls "make _report" and not "make report".  Working on a proposed fix.

markrtuttle (approved):

PR #68, opened 2021-09-08
Change license from Apache-2.0 to MIT-0

markrtuttle:
This pull request changes the starter kit license from Apache-2.0 to MIT-0.  This pull request changes a lot of files, and the changes are best reviewed commit-by-commit:

* First commit changes the LICENSE file
* Second commit changes the SPDX-License-Identifier in 50+ files (the same change in all files)
* Third commit removes SPDX-License-Identifier from files copied from the starter kit into the project repository
* Fourth commit contributes a script for use with an existing project repository: The script removes SPDX-License-Identifier from all files copied from the starter kit into the project repository.


hyandell:
No concern with the relicensing itself.

The script is a thoughtful touch, but it feels like overkill. If you ran it on an Apache-2.0 licensed project from Amazon, it's going to remove a lot more source headers than intended (I think)?

hyandell (commented):
No concern with the relicensing itself.

The script is a thoughtful touch, but it feels like overkill. If you ran it on an Apache-2.0 licensed project from Amazon, it's going to remove a lot more source headers than intended (I think)?

remove-apache-license-identifier.py
```diff
@@ -0,0 +1,184 @@
...
+
+def find_apache_references(proofdir=None):
+    paths = run(['git', 'grep', '-l', 'Apache', '.'], cwd=proofdir)
+    paths = [os.path.normpath(os.path.join(proofdir, path)) for path in paths]
+    paths = [path for path in paths if not os.path.islink(path)]
+    return sorted(paths)
+
+def remove_apache_reference(path, extension='backup'):
+    apache_references = [
+        '# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.',
```
karkhaz, line 128: Why are we removing this line? I would assume that we only want to get rid of the SPDX line below, not the copyright declaration.

Also, why are we deleting lines? Isn't the point of this script to change Apache into MIT (which is what the rest of this PR actually does)?

I wonder whether it might be more convenient for this script to just replace `Apache-2.0` in the relevant files with an SPDX string of the user's choice, defaulting to `MIT-0`?

karkhaz:
@hyandell the script contains a list of specific filenames to modify (and furthermore also checks that they are not symbolic links to a submodule, in which case we wouldn't want to modify the license). Also, it looks like this script is intended to be run from the subdirectory containing the files we'd want to modify, rather than the root of the repo. So it shouldn't touch the rest of the Amazon project.

remove-apache-license-identifier.py
markrtuttle: This is the proposal worked out in discussion with AWS IoT and AWS Open Source.

markrtuttle:
@hyandell The script is only interesting to Amazon project that have already built proofs on top of the starter kit, and have to retrofit to be consistent with this version of the starter kit.  I don't mind deleting it and circulating the script to the appropriate projects via another path.  Let me know.

karkhaz (approved):

hyandell (approved):
Looks good to me :)

I totally missed that the files were defined in copied_files. I've no concerns with the script with that filter being there.

PR #69, opened 2021-09-08
Add Litani job to check for function declarations

karkhaz:
This commit adds a job that ensures that the link job does not complain
about missing function declarations. It checks for a certain warning in
the output of the link job; if this warning is printed, that indicates
that there is no declaration of the function under test in the proof.
This is a user error, so this commit ensures that the proof is aborted
if this happens.

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -491,18 +491,28 @@ $(HARNESS_GOTO)1.goto: $(PROOF_GOTO)1.goto $(PROJECT_GOTO)3.goto
 	$(LITANI) add-job \
 	  --command '$(GOTO_CC) $(CBMC_VERBOSITY) --function $(HARNESS_ENTRY) $^ $(LINK_FLAGS) -o $@' \
 	  --inputs $^ \
-	  --outputs $@ \
+	  --outputs $@ $(LOGDIR)/link_proof_project-log.txt \
 	  --stdout-file $(LOGDIR)/link_proof_project-log.txt \
 	  --interleave-stdout-stderr \
 	  --pipeline-name "$(PROOF_UID)" \
 	  --ci-stage build \
 	  --description "$(PROOF_UID): linking project to proof"
 
+.PHONY: $(PROOF_UID)_link_check
+$(PROOF_UID)_link_check: $(HARNESS_GOTO)1.goto
+	$(LITANI) add-job \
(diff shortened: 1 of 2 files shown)


karkhaz:
[Here](https://d39z7xfr4hj3jh.cloudfront.net/final/d6ec5990-a229-4dfd-a9b9-0fa465e315e1/pipelines/httpParserOnHeadersCompleteCallback/index.html#job-6932345e-16cc-4ac6-8391-f3ddb5675a64)'s an example of an error that this catches.

markrtuttle:
I'm reluctant to merge this into the starter kit.  This is catching a careless mistake, but is it a soundness bug?

If I have foo.c 
```
int foo(int x) {
  return x+1;
}
```
and bar.c
```
#include <assert.h>

int main() {
  int x = foo(1);
  assert(x == 2);

}
```
and I do
```
goto-cc -o foo.o -Wall foo.c
goto-cc -o bar.o -Wall bar.c
goto-cc -o main foo.o bar.o
cbmc main
```
I get
```
bar.c: In function 'main':
bar.c:4:1: warning: function 'foo' is not declared
   int x = foo(1);
   
CBMC version 5.38.0 (cbmc-5.38.0) 64-bit x86_64 macos
VERIFICATION SUCCESSFUL
```
and failure if I change the assert to x == 1.  If I look at the goto binary, I see the expected definition of foo and main, and I see the expected invocation of foo from main.

karkhaz:
This is really intended to help catch the specific case where the user didn't correctly declare the mangled name of a static function, which you had mentioned resulted in problems with CBMC Viewer.

It's not a soundness bug. But I'm also thinking about the case of undeclared functions, which potentially is. You're doing an audit for such functions right now but we have no mechanism for preventing more from being introduced in the future. It would be good to have a script that checks the output of all goto-cc and cbmc invocations for certain warnings, not just this one. I intended this script to be easily extensible with additional warnings to check for, so that we can add to it every time we notice users making errors in their proof harnesses.

markrtuttle:
I am modifying viewer to summarize the results of the audit in json and render the results in the reports for individual proofs.  I would like Wellspring to grovel over those json files and summarize the results on the top level report (I have the script for that) and alarm based on that if necessary.

karkhaz:
Closing as per offline discussions, we're going to implement this in collaboration with the JSON output from CBMC Viewer.

PR #72, opened 2021-09-28
Quiet noisy output from Makefile.common

markrtuttle:
This patch suppresses the noisy output of Makefile.common (mostly the echoing of the long commands used to invoke 'litani add-job' used to create the ninja build graph) unless the Makefile variable VERBOSE is defined (eg, with make VERBOSE=1 report).

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -351,6 +351,11 @@ CBMC_USE_FUNCTION_CONTRACTS := $(patsubst %,--replace-call-with-contract %, $(US
 # the APPLY_LOOP_CONTRACTS variable to 1.
 APPLY_LOOP_CONTRACTS ?= 0
 
+# Silence makefile output (eg, long litani commands) unless VERBOSE is set.
+ifndef VERBOSE
+    MAKEFLAGS := $(MAKEFLAGS) -s
+endif
+
 ################################################################
 ################################################################
 ## Section II: This section is for project-specific definitions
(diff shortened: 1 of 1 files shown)


feliperodri (approved):
LGTM

danielsn (approved):

template-for-repository/proofs/Makefile.common
```diff
@@ -351,6 +351,14 @@ CBMC_USE_FUNCTION_CONTRACTS := $(patsubst %,--replace-call-with-contract %, $(US
 # the APPLY_LOOP_CONTRACTS variable to 1.
 APPLY_LOOP_CONTRACTS ?= 0
 
+# Set VERBOSE=1 to see the full Makefile output, but silence by default
+# the noisy output (for example, the long `litani add-job` invocations)
+ifdef VERBOSE
+  QUIET =
```
karkhaz, line 357: Instead of decorating every single recipe, why not just pass the `-q` flag to Make?

```
ifndef VERBOSE
    MAKEFLAGS := $(MAKEFLAGS) -q
endif
```
I think that should be sufficient

karkhaz (changes requested):

markrtuttle:
I've modified the Makefile as Kareem recommended (adding the `-s` flag to MAKEFLAGS).  I have retained the progress information printed by long-running targets: On slow machines, it is comforting to see progress information printed during the four or five seconds before ninja starts printing its own progress information.

template-for-repository/proofs/Makefile.common
markrtuttle: Done

karkhaz (approved):

PR #77, opened 2021-10-18
Reorder CBMC passes as required for code contracts

SaswatPadhi:
*Issue #, if available:*
n/a

*Description of changes:*

Loop unwinding and contract replacement were previously performed _after_ function contract enforcement. There was no dependence between these passes before. However, now they must be done _before_ function contract enforcement:
- Loop unwinding must be done before enforcing function contracts, since loop unwinding does not replicate writeset snapshots that might be introduced during contract enforcement step.
- Function call replacement using contracts must be done before enforcing function contracts, since contract enforcement step inlines all function calls.
- Loop contract application must be done before enforcing function contracts.

Note that since loop unwinding is now done much earlier, the loop "numbers" used in proof `Makefile`s might need to be changed.

A second commit further reorders the slicing passes to reduce the code size on which we do contract instrumentation. In particular, the `--apply-loop-contracts` flag instruments _ALL_ loops that have contracts, even those unreachable from the entry point. This is unnecessary since unreachable loops were sliced away in the next step. After this change, they would be sliced away early on, and only the loops reachable from the entrypoint function would be instrumented.

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -507,10 +507,59 @@ else
 	  --description "$(PROOF_UID): restricting function pointers in project sources"
 endif
 
+# Optionally fill static variable with unconstrained values
+$(HARNESS_GOTO)3.goto: $(HARNESS_GOTO)2.goto
+ifeq ($(NONDET_STATIC),"")
+	$(LITANI) add-job \
+	  --command 'cp $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ \
+	  --pipeline-name "$(PROOF_UID)" \
+	  --ci-stage build \
+	  --description "$(PROOF_UID): not setting static variables to nondet"
+else
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) $(NONDET_STATIC) $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ \
+	  --stdout-file $(LOGDIR)/nondet_static-log.txt \
+	  --interleave-stdout-stderr \
+	  --pipeline-name "$(PROOF_UID)" \
+	  --ci-stage build \
+	  --description "$(PROOF_UID): setting static variables to nondet"
+endif
+
+# Omit unused functions (sharpens coverage calculations)
+$(HARNESS_GOTO)4.goto: $(HARNESS_GOTO)3.goto
+	$(LITANI) add-job \
+	  --command \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) --drop-unused-functions $^ $@' \
+	  --inputs $^ \
+	  --outputs $@ \
+	  --stdout-file $(LOGDIR)/drop_unused_functions-log.txt \
+	  --interleave-stdout-stderr \
+	  --pipeline-name "$(PROOF_UID)" \
+	  --ci-stage build \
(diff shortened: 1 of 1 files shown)


template-for-repository/proofs/Makefile.common
```diff
@@ -119,7 +119,7 @@ PROOFDIR ?= $(abspath .)
 CBMCDIR ?= $(PROOF_ROOT)/cbmc
 
 # Default CBMC flags used for property checking and coverage checking
-CBMCFLAGS += --unwind 1 $(CBMC_UNWINDSET) --flush
+CBMCFLAGS += --unwind 1 --flush
```
feliperodri, line 122: We should add ac comment explaining why we need `--unwind 1` here.

feliperodri (approved):

template-for-repository/proofs/Makefile.common
SaswatPadhi: Good idea; updated.

SaswatPadhi:
The loop numbers need to be updated in proof `Makefile`s now.
Draft PRs on [s2n-tls](https://github.com/aws/s2n-tls) and [aws-c-common](https://github.com/awslabs/aws-c-common):
- https://github.com/aws/s2n-tls/pull/3096
- https://github.com/awslabs/aws-c-common/pull/855

SaswatPadhi:
I guess we should discuss this change a bit more before merging.

Martin says (https://github.com/diffblue/cbmc/issues/6432#issuecomment-959405490) that the `--unwind` semantics on `cbmc` (which is what we currently use) is not the same as `--unwind` semantics on `goto-instrument` (which is what this PR proposes).

SaswatPadhi:
Per offline discussion on 11/10/2021 with @feliperodri and @tautschnig -- When we switch to unwinding with `goto-instrument`, we wouldn't need the `+1` that we are currently doing because we pass `--unwind` to `cbmc`, but there's also no harm in unwinding "more" so we could just leave it that way in all our `Makefile`s.

feliperodri:
> The loop numbers need to be updated in proof `Makefile`s now. Draft PRs on [s2n-tls](https://github.com/aws/s2n-tls) and [aws-c-common](https://github.com/awslabs/aws-c-common):
> 
>     * [Bump aws-templates-for-cbmc-proofs aws/s2n-tls#3096](https://github.com/aws/s2n-tls/pull/3096)
> 
>     * [Bump aws-template-for-cbmc-proofs aws-c-common#855](https://github.com/awslabs/aws-c-common/pull/855)

Is this the only blocker for this PR? @SaswatPadhi

SaswatPadhi:
Yes, and I have fixed the proofs for both these repos -- only changes to loop numbers was necessary. I can push the fixes in a bit, but similar loop-numbers changes will be necessary in all repos that use starter kit when we move the submodule pointer forward.

SaswatPadhi:
If there are other important PRs that need to go it first, then we could merge those first and move starter kit pointer forward. Then merge this and move the pointer forward along with several changes to loop numbers.

feliperodri (approved):

markrtuttle:
I've added a third commit [688d5d5e](https://github.com/awslabs/aws-templates-for-cbmc-proofs/pull/77/commits/688d5d5ed4d5b0942a3c4ed360ec46085a3fe256) that restores loop unwinding to the invocation of cbmc and added a comment to the invocation of goto-instrument doing the loop unwinding explaining what is going on.

feliperodri (approved):

PR #78, opened 2021-10-19
Cleanup make output in run-cbmc-proofs.py

markrtuttle:
This pull request adds `strip_make_output` to `run-cbmc-proofs.py` to clean up the standard output of make when invoked by subprocess.  The progress information being deleted was observed when trying to write another script using subprocess to invoke `run-cbmc-proofs.py` itself.


karkhaz:
I just had a cursory glance at [cbmc-viewer #57](https://github.com/awslabs/aws-viewer-for-cbmc/pull/57) and it definitely seems like you could use `--no-standalone` to great advantage there.

I might call you tomorrow to pick your brains about exactly what you're doing, but: it seems like you're running Litani, and then doing a 'post-processing' phase using the finished run.json. What I suggest instead is to use Litani itself to run the tests. That way you don't have to wait for every single HTTP proof to finish before commencing the viewer test run: instead, each viewer test will start as soon as that one proof finishes.

The idea is that the `report` job of each proof should specify that it emits the 'viewer.json` (and other files) as outputs, using the `--output` flag to Litani. Then you add some additional Litani jobs, which take those JSON files as inputs: those are your test suite. Then you run the whole lot. This is one of the many cases that the `--no-standalone` flag was intended for, to extend your proof run with arbitrary additional jobs that Litani can run for you.

PR #80, opened 2021-11-04
Restrict function pointers in linked code, not just project code

markrtuttle:
This commit modifies Makefile.common to instrument the linked goto program with function pointer restrictions, and not just the project code before linking with the proof code.  The problem was when the function pointer being restricted was in the project code and the stub for the function was in the proof code.


SaswatPadhi (approved):
LGTM

danielsn (approved):

karkhaz (approved):

PR #84, opened 2021-12-06
Apply --object-bits during goto-cc compilation

feliperodri:
Signed-off-by: Felipe R. Monteiro felisous@amazon.com

*Issue #, if available:*
N/A.

*Description of changes:*

We must ensure we use the same object size during compilation and CBMC invocation.

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -402,6 +402,7 @@ COMMA :=,
 # Set C compiler defines
 
 CBMCFLAGS += --object-bits $(CBMC_OBJECT_BITS)
+COMPILE_FLAGS += --object-bits $(CBMC_OBJECT_BITS)
 
 DEFINES += -DCBMC=1
 DEFINES += -DCBMC_OBJECT_BITS=$(CBMC_OBJECT_BITS)


template-for-repository/proofs/Makefile.common
```diff
@@ -402,6 +402,7 @@ COMMA :=,
 # Set C compiler defines
 
 CBMCFLAGS += --object-bits $(CBMC_OBJECT_BITS)
```
remi-delmas-3000, line 404: Does `--object-bits` need to be passed to both `goto-cc` and `cbmc` or is is sufficient to pass it only to `goto-cc`? It may be safer to use this parameter only once.

remi-delmas-3000 (changes requested):

markrtuttle (approved):

template-for-repository/proofs/Makefile.common
feliperodri: Good point, @remi-delmas-3000! I'd expect that we only need to pass it during compilation with `goto-cc`. @tautschnig any insights on this?

feliperodri: @remi-delmas-3000 based on an offline discussion, it appears we must pass the same value for both invocations.

remi-delmas-3000 (approved):

tautschnig (approved):
I am, however, wondering how we managed to go this long without having this in place. Has every actual project already made such adjustments?

PR #89, opened 2022-02-03
Set value for LITANI env. variable during setup

angelonakos:
*Issue #, if available:*

*Description of changes:*

When a user is setting up the CBMC starter kit to verify their project, the value of the `LITANI` environment variable
will now be set to `litani` if Litani can be found in path.

### DIFF
--- a/scripts/setup.py
+++ b/scripts/setup.py
@@ -8,6 +8,7 @@
 from pathlib import Path
 import logging
 import os
+import shutil
 
 import repository
 import util
@@ -21,7 +22,7 @@ SRCDIR ?= $(abspath $(PROOF_ROOT)/{})
 LITANI_TEXT = """
 # Absolute path to the litani script.
 #
-LITANI ?= $(abspath $(PROOF_ROOT)/{})
+LITANI ?= {}
 """
 
 PROJECT_TEXT = """
@@ -45,8 +46,7 @@ def create_makefile_template_defines(
     with open(makefile, "w", encoding='utf-8') as fileobj:
         print(SRCDIR_TEXT.format(os.path.relpath(source_root, proof_root)),
               file=fileobj)
-        print(LITANI_TEXT.format(os.path.relpath(litani, proof_root)),
-              file=fileobj)
+        print(LITANI_TEXT.format(litani), file=fileobj)
         print(PROJECT_TEXT.format(project_name), file=fileobj)
 
 def main():
@@ -57,7 +57,11 @@ def main():
     cbmc_root = Path.cwd()
     proof_root = cbmc_root / "proofs"
     source_root = repository.repository_root()
-    litani = repository.litani_root() / 'litani'
+    litani = "litani" if shutil.which("litani") else \
+        repository.litani_root() / "litani"
+    if litani != "litani":
+        relpath_from_litani_to_proof_root = os.path.relpath(litani, proof_root)
+        litani = f"$(abspath $(PROOF_ROOT)/{relpath_from_litani_to_proof_root})"
     project_name = util.ask_for_project_name()
 
     util.copy_repository_templates(cbmc_root)


karkhaz:
Thanks! I think we should continue to support the use-case of having a litani submodule. That means a few things...
* the run-script shouldn't hard-code the path to litani. Actually, it's pretty annoying that we write the path twice, once in the makefile and once in the run script.
  * Can I suggest that you add a new rule to `Makefile.common`, similar to `echo-proof-uid` and `echo-project-name`, that echoes the Litani path? Then the run script could run `make echo-litani-path` to get the path that is written in `Makefile-template-defines`.
* Can the setup script give the user a choice of whether to submodule or use an installed version of Litani? I think a reasonably verbose dialogue would be appropriate here---explain to the user that they can either distribute Litani with the repository (if the user chooses this, then ask for a path and run `git submodule add` for them) or they can assume that Litani is already installed (this is the recommended option, users who want to run profs can run `brew install ...`), etc.

So in summary, the setup script will write the path to Litani just once to `Makefile-template-defines` (either just `litani`, or the path to the submodule), and the run script should use that rather than hardcoding it.

scripts/setup.py
```diff
@@ -43,7 +43,13 @@ def create_makefile_template_defines(
...
         print(SRCDIR_TEXT.format(os.path.relpath(source_root, proof_root)),
               file=fileobj)
-        print(LITANI_TEXT.format(os.path.relpath(litani, proof_root)),
+        if is_litani_installed:
+            litani_path = '"litani"'
+        else:
+            litani_path = util.read_litani_path()
+            relpath = os.path.relpath(litani_path, proof_root)
+            litani_path = f"$(abspath $(PROOF_ROOT)/{relpath})"
+        print(LITANI_TEXT.format(litani_path),
```
nrdg42, line 52: I would remove the `is_litani_installed` parameter from the function and instead create a utils function called `get_litani_path()` which would do something like:

```python
def get_litani_path():
    if util.is_litani_installed():
        return pathlib.Path("litani")
    return util.read_litani_path()
```
that way the code here would be:
```python
litani_path = util.get_litani_path().relative_to(proof_root)
```
Also do we want to use the `os` or the `pathlib` libraries for paths? I'm more familiar with `pathlib` but it seems as though we're using `os` in the other script so you'll probably need to convert .

nrdg42: Actually nevermind I don't think that would work how I thought it would

angelonakos: The `is_litani_installed` parameter has been removed.

angelonakos: > Also do we want to use the `os` or the `pathlib` libraries for paths?

I'm open to making that change. However, it should not be part of this PR as it points to a need to conduct a broader minor refactoring.

nrdg42 (approved):

scripts/setup.py
```diff
@@ -60,8 +60,13 @@ def main():
...
 
-    # the script is linking to the litani script within the litani submodule
-    litani = util.read_litani_path()
+    # the script is linking to litani
+    if util.is_litani_installed():
+        litani = '"litani"'
+    else:
+        litani_path = util.read_litani_path()
+        relpath = os.path.relpath(litani_path, proof_root)
+        litani = f"$(abspath $(PROOF_ROOT)/{relpath})"
```
nrdg42, line 69: You're still passing `litani` as a parameter to the function which I'm okay with because that is how it was before. I personally think this logic could just be handled within the function instead of being passed in as a parameter, so I'm curious as to if there's a reason its being done this way, but I'm not going to block on it

angelonakos: I just abstracted this logic inside its own function as pointed out by Nick

scripts/util.py
```diff
@@ -51,8 +52,17 @@ def read_source_root_path():
...
 
+def is_litani_installed():
+    dialogue = textwrap.fill("Litani can either be distributed with your " \
+        "repository in the form of a submodule, in which case you will need " \
+        "to provide the desired directory, or it can be installed using: " \
+        "brew install aws/tap/litani")
+    print(dialogue)
+    print("Has litani already been installed in your system? [Y/n]", end="")
+    return read_from_stdin().lower() in ("y", "yes")
+
```
ronakfof, line 63: Will this be better?
```
def is_litani_installed():
    dialogue = textwrap.fill("Litani can either be distributed with your " \
        "repository in the form of a submodule, in which case you will need " \
        "to provide the desired directory, or it can be installed using: " \
        "brew install aws/tap/litani")
    print(dialogue)
    if shutil.which("litani"):
        LITANI_VERSION = subprocess.run(
            "litani -V",
            shell=True,
            capture_output=True).stdout.decode("utf-8").rstrip()
        print(f"litani-{LITANI_VERSION} Installation found")
    print("Do you wanna use this installation of Litani? [Y/n]", end="")
    return read_from_stdin().lower() in ("y", "yes")
```

angelonakos: I incorporated Ronak's suggestion with a small change of my own

ronakfof (approved):

scripts/util.py
```diff
@@ -51,8 +53,28 @@ def read_source_root_path():
...
+    print(textwrap.fill("Litani can either be distributed with your " \
+        "repository in the form of a submodule, in which case you will need " \
+        "to provide the desired directory, or it can be installed using: " \
+        "brew install aws/tap/litani"))
+    if not shutil.which("litani"):
+        print(textwrap.fill("Litani is not installed. Either exit this CLI, " \
+            "install Litani using brew, and run the script again or proceed " \
+            "with the option of including Litani as a submodule"))
+        return False
+    litani_version = subprocess.run(
```
karkhaz, line 66: I don't think there's any need for this part of the dialogue. If litani is installed with brew, then the version number will change anyway. I don't think there's any point in making the user commit to a particular version number.

karkhaz (changes requested):

karkhaz:
Can this be two commits? One refactoring the run script to use the value of LITANI from the Makefile, and the second one to set that value with the new dialog?

template-for-repository/proofs/Makefile.common
```diff
@@ -938,6 +938,10 @@ echo-proof-uid:
 echo-project-name:
 	@echo $(PROJECT_NAME)
 
+.PHONY: echo-litani-path
+echo-litani-path:
```
karkhaz, line 942: It actually looks like there's already a rule called `litani-path` that does this? No need for this change. 🧲

template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -197,13 +197,12 @@ def run_build(litani, jobs):
...
 
-def get_litani_path(proof_root):
+def get_litani_path():
     cmd = [
         "make",
         "--no-print-directory",
-        "PROOF_ROOT=%s" % proof_root,
         "-f", "Makefile.common",
-        "litani-path",
+        "echo-litani-path",
```
karkhaz, line 205: 🧲 No need.

karkhaz (changes requested):
Sorry to ask for a change again, but I think this is still too much text to confront the user with. Can we do:
* If `shutil.which("litani")` then just use `litani` and don't ask the user anything at all.
* Else:
```
You need to install Litani or submodule it into this codebase to
run CBMC proofs. Choose an option:
1. Run `brew install litani` to install Litani on this machine (RECOMMENDED)
2. Submodule Litani into this codebase
3. Do nothing, manually specify path to Litani later
Enter your choice [1,2,3]:
```
If 1 or empty, run `brew install litani` and use `"litani"` for the value.
If 3, use `"litani"` for the value.
If 2, ask a second question: `Litani will now be submoduled into this repository. Enter the path to the directory that Litani should be submoduled into:`; submodule Litani into that directory and use that path.

angelonakos:
> Sorry to ask for a change again, but I think this is still too much text to confront the user with. Can we do:
> 
> * If `shutil.which("litani")` then just use `litani` and don't ask the user anything at all.
> 
> * Else:
> 
> ```
> 
> You need to install Litani or submodule it into this codebase to
> 
> run CBMC proofs. Choose an option:
> 
> 1. Run `brew install litani` to install Litani on this machine (RECOMMENDED)
> 
> 2. Submodule Litani into this codebase
> 
> 3. Do nothing, manually specify path to Litani later
> 
> Enter your choice [1,2,3]:
> 
> ```
> 
> If 1 or empty, run `brew install litani` and use `"litani"` for the value.
> 
> If 3, use `"litani"` for the value.
> 
> If 2, ask a second question: `Litani will now be submoduled into this repository. Enter the path to the directory that Litani should be submoduled into:`; submodule Litani into that directory and use that path.

I have implemented the above options as suggested by Kareem.

The changes can now be reviewed.

scripts/util.py
```diff
@@ -51,8 +64,77 @@ def read_source_root_path():
...
+    try:
+        sys.path.insert(1, path)
+        importlib.import_module("litani")
+        is_litani_submodule_in_path = True
+    except ModuleNotFoundError:
+        print(f"Litani submodule was not found at directory: {path})")
+        is_litani_submodule_in_path = False
+    finally:
+        del sys.path[1]
+    return is_litani_submodule_in_path
```
nrdg42, line 77: Personal style preference but not blocking. Also you are overwriting the function name with the boolean name `is_litani_submodule_in_path` which probably doesn't cause any issues but is something to be aware of I guess.

```python
    is_in_path = False
    sys.path.insert(1, path)
    try:
        importlib.import_module("litani")
        is_in_path = True
    except ModuleNotFoundError:
        print(f"Litani submodule was not found at directory: {path})")
    del sys.path[1]
    return is_in_path
```

scripts/util.py
```diff
@@ -51,8 +64,77 @@ def read_source_root_path():
...
+def determine_litani_origin():
+    choice = None
+    is_litani_option_chosen = False
+    while not is_litani_option_chosen:
+        print(LITANI_DIALOG, file=sys.stderr)
+        choice = read_from_stdin()
+        if choice not in ("1", "2", "3"):
+            print("You have entered an invalid value. Please choose again.")
+        else:
+            is_litani_option_chosen = True
```
nrdg42, line 101: Again, personal style preference, not blocking:
```python
def determine_litani_origin():
    choice = None
    is_litani_option_chosen = False
    while True:
        print(LITANI_DIALOG, file=sys.stderr)
        choice = read_from_stdin()
        if choice in ("1", "2", "3"):
            break
        print("You have entered an invalid value. Please choose again.")
    return choice
```

scripts/util.py
```diff
@@ -51,8 +64,77 @@ def read_source_root_path():
...
+    except subprocess.CalledProcessError as e:
+        logging.warning("Failed in running `brew install litani`")
+        logging.warning(e)
+
+def install_litani_submodule(relpath):
+    absolute_path = os.path.abspath(os.path.expanduser(relpath))
+    if not is_litani_submodule_in_path(absolute_path):
+        print(textwrap.fill(
+            "Do you want the Litani submodule to be added to the "
+            "specified path? [y/Y] "), file=sys.stderr)
```
nrdg42, line 122: Usually the dialogue is `[y/N]` or `[Y/n]` I believe, with the recommended/default choice being the capital, not `[y/Y]`

angelonakos: Thanks. Switched to [Y/n]. Hadn't noticed the typographical error.

scripts/util.py
angelonakos: Switched to this

scripts/util.py
angelonakos: Switched to almost this (kept insertion statement inside of try-except block).

scripts/util.py
```diff
@@ -3,9 +3,14 @@
 
 """Methods of manipulating the templates repository."""
 
+from asyncio.subprocess import STDOUT
```
karkhaz, line 6: Can we not do this? I don't like that `STDOUT` somewhere far down in the codebase actually refers to something in the asyncio library, it's just easier to read if this is made clear at the use site IMHO

scripts/util.py
```diff
@@ -51,8 +64,74 @@ def read_source_root_path():
 def read_proof_root_path():
     return read_path_from_stdin("the 'proofs' directory (usually '.')")
 
-def read_litani_path():
-    return read_path_from_stdin("the litani script")
+def is_litani_submodule_in_path(path):
+    is_in_path = False
+    try:
+        sys.path.insert(1, path)
```
karkhaz, line 70: I'd really rather not couple this script to how the litani source code is arranged. Can I suggest that you just check that there's a directory called either `litani` or `aws-build-accumulator` at the path? No need to look inside I think.

scripts/util.py
```diff
@@ -51,8 +64,74 @@ def read_source_root_path():
...
+def is_litani_installed():
+    return shutil.which("litani") is not None
+
+def determine_litani_origin():
+    choice = None
+    while True:
+        print(LITANI_DIALOG, file=sys.stderr)
+        choice = read_from_stdin()
+        if choice in ("1", "2", "3"):
+            break
```
karkhaz, line 97: `return choice`? Or even `return int(choice)`?

scripts/util.py
```diff
@@ -51,8 +64,74 @@ def read_source_root_path():
...
+        print("You have entered an invalid value. Please choose again.")
```
karkhaz, line 98: stderr?

scripts/util.py
```diff
@@ -51,8 +64,74 @@ def read_source_root_path():
...
+             "https://github.com/awslabs/aws-build-accumulator.git", "litani"],
+             check=True, cwd=path)
+    except subprocess.CalledProcessError:
+        print(f"Could not add Litani submodule at path: {path}")
+
+def is_litani_installed():
+    return shutil.which("litani") is not None
+
+def determine_litani_origin():
+    choice = None
```
karkhaz, line 92: No need

scripts/util.py
```diff
@@ -51,8 +64,74 @@ def read_source_root_path():
...
+            break
+        print("You have entered an invalid value. Please choose again.")
+    return choice
+
+def install_litani_brew_package():
+    """Runs `brew install litani`. Exits if a failure occurs"""
+    try:
+        proc = subprocess.run(
+            ["brew", "install", "litani"],
+            check=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
karkhaz, line 106: Can I suggest that we don't buffer stderr here? I think it's more reassuring for the user to see it, whether there's an error or not. I think that might even be true of stdout. No need to hide it. Below, just do `sys.exit(1)` without logging.

(Even if you don't want to do what I suggest above, there's no need to do `decode()` below. Just pass `text=True` to `subprocess.run`)

scripts/util.py
```diff
@@ -51,8 +64,74 @@ def read_source_root_path():
...
+        if proc.returncode:
+            logging.warning(proc.stderr.strip().decode())
+            sys.exit(1)
+    except subprocess.CalledProcessError as e:
```
karkhaz, line 110: This isn't reachable because you're doing `check=False` I think?

What I actually suggest is:
1. `check=True`
2. Don't bother with `if proc.returncode`
3. Catch both `CalledProcessError`, and also `FileNotFoundError` (I think this gets raised if `brew` is not on the PATH or not executable) and handle the error there. If brew fails I think its own stderr should be sufficient, if it's not on the PATH then logging a message would probably helpful.

scripts/util.py
```diff
@@ -23,6 +28,14 @@
     ".gitignore",
 ]
 
+LITANI_DIALOG = textwrap.dedent("""\
+    You need to install Litani or submodule it into this codebase to
+    run CBMC proofs. Choose an option:
+    1. Run `brew install litani` to install Litani on this machine (RECOMMENDED)
```
karkhaz, line 34: 🌮

scripts/util.py
```diff
@@ -51,8 +64,74 @@ def read_source_root_path():
...
+    absolute_path = os.path.abspath(os.path.expanduser(relpath))
+    if not is_litani_submodule_in_path(absolute_path):
+        print(textwrap.fill(
+            "Do you want the Litani submodule to be added to the "
+            "specified path? [Y/n] "), file=sys.stderr)
+        if read_from_stdin().lower() in ("y", "yes"):
+            add_litani_submodule(relpath)
+    return f"$(abspath $(PROOF_ROOT)/{relpath})"
+
+def get_litani():
```
karkhaz, line 124: I don't like how decoupled this logic is from the actual messages to the user.

Can the text in 🌮 be something like:
```
print("choose an option: ")
for idx, choice in INSTALL_CHOICES.items():
  print(f"{idx}. {choice['description']}")
```
`INSTALL_CHOICES` is
```
{
  "1": {
     "description": "Run `brew install litani`",
     "action": install_litani_brew_package,
  }
```
Then this method is basically
```
return INSTALL_CHOICES[determine_litani_origin()]["action"]()
```

karkhaz (changes requested):

scripts/util.py
angelonakos: Yes. Will remove. This doesn't belong here. I accidentally allowed the IDE to add this import statement, when I added `stdout` some place else.

scripts/util.py
angelonakos: In the case that a user has already installed Litani using brew, removing `stderr=subprocess.PIPE` will result into the following message popping us (written to stderr). The message:

```
Warning: litani 1.22.0 is already installed and up-to-date.
To reinstall 1.22.0, run:
  brew reinstall litani
```
We're fine with that, right? I personally am fine with this message popping up and with removing the `stderr` as you pointed out.

scripts/util.py
angelonakos: wrt 3.:

We indeed have to handle for `FileNotFoundError` as well. Thanks for pointing that out.

Informative messages regarding the underlying error will be logged out even if we just log the error itself.

For example, if I change this function to instead attempt to install "slitani" via brew (intended typographical error) then I get:

```
Warning: No available formula with the name "slitani". Did you mean litani?
Error: No previously deleted formula found.
==> Searching taps on GitHub...
Error: No formulae found in taps.
WARNING:root:Failed in running `brew install litani`
WARNING:root:Command '['brew', 'install', 'slitani']' returned non-zero exit status 1.
```

Additionally, if I update my `PATH` and make `brew` temporarily unavailable then calling the soon-to-be-revised function that logs the error then you would see:

```
WARNING:root:`brew` is missing from your PATH
WARNING:root:[Errno 2] No such file or directory: 'brew'
```

which itself should be informative enough.

scripts/util.py
karkhaz: I think this is fine, but also this method doesn't even get called if Litani was already on the $PATH, right?

scripts/util.py
angelonakos: This looks much clearer. I can add the `INSTALL_CHOICES` dictionary, however, it will only be a key-value pair of index to description. The reason being that the action for index 2. (i.e. install Litani using the submodule approach) requires a parameter to be passed. This requirement breaks the neat one-liner `return` statement: `return INSTALL_CHOICES[determine_litani_origin()]["action"]()` since I wouldn't be able to pass the `proof_root` argument that is needed by the second action

ronakfof (approved):

markrtuttle (approved):

scripts/setup.py
```diff
@@ -57,7 +57,11 @@ def main():
     cbmc_root = Path.cwd()
     proof_root = cbmc_root / "proofs"
     source_root = repository.repository_root()
-    litani = repository.litani_root() / 'litani'
+    litani = "litani" if shutil.which("litani") else \
+        repository.litani_root() / "litani"
+    if litani != "litani":
```
karkhaz, line 62: The condition seems incorrect.
```
>>> shutil.which("litani")
'/usr/local/bin/litani'
```

karkhaz (changes requested):

PR #90, opened 2022-03-08
Adjust unwinding flags for cbmc and goto-instrument

SaswatPadhi:
*Issue #, if available:*

n/a

*Description of changes:*

When loop contracts are instrumented, we need to raise the unwinding value (from the default `1` value) to at least `2` to allow for unwinding the instrumentation on both the base case and the step case.
This is because, the new instrumentation routine [diffblue/cbmc/pull/6701] generates a back jump to go over the same loop twice.

I have tested the changes on all of the CoreJSON proofs with the latest build of CBMC from `develop` branch.

### DIFF
--- a/template-for-repository/proofs/Makefile.common
+++ b/template-for-repository/proofs/Makefile.common
@@ -119,8 +119,7 @@ PROOFDIR ?= $(abspath .)
 CBMCDIR ?= $(PROOF_ROOT)/cbmc
 
 # Default CBMC flags used for property checking and coverage checking.
-UNWIND_COUNT = 1
-CBMCFLAGS += --unwind $(UNWIND_COUNT) $(CBMC_UNWINDSET) --flush
+CBMCFLAGS += --flush
 
 # Do property checking with the external SAT solver given by
 # EXTERNAL_SAT_SOLVER.  Do coverage checking with the default solver,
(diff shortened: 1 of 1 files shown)


template-for-repository/proofs/Makefile.common
```diff
@@ -119,7 +119,8 @@ PROOFDIR ?= $(abspath .)
 CBMCDIR ?= $(PROOF_ROOT)/cbmc
 
 # Default CBMC flags used for property checking and coverage checking.
-CBMCFLAGS += --unwind 1 $(CBMC_UNWINDSET) --flush
+UNWIND_COUNT = 1
```
karkhaz, line 122: Shouldn't this be `?=` like the other variables so that it can be overridden in each project or in particular proofs?

feliperodri (changes requested):

template-for-repository/proofs/Makefile.common
SaswatPadhi: Hi Felipe, Kareem,

This was hardcoded and not exposed before, so I didn't want to expose it in this PR. But thinking more about whether this should be exposed to end users, I think it shouldn't be, for several reasons:

1. `UNWINDSET` is exposed for users to indicate unwinding counts on individual loops. This is just the default `UNWIND_COUNT` for all the remaining loops that are not in the `UNWINDSET`.  
Setting it to a small value makes CBMC fail fast when the user forgets to add a particular to `UNWINDSET`. Users might potentially shoot themselves in the foot if they start tweaking this default value instead -- proofs might take way too long only to fail on an unwinding assertion violation.
2. Suppose the user sets this to `1`, but if they are also using loop contracts, then we would override their choice and set it to `2`, or we would error out saying that loop contracts require `UNWIND_COUNT` of at least 2.

I'm happy to change to `=` to `?=`, but I don't yet see significant value in exposing this. The `UNWINDSET` is what the users should tweak.

What are your thoughts?

markrtuttle: Could we have
```
UNWIND ?= --unwind 1
```
just as we do for other CBMC flags like `--bounds-check`, and then later have
```
$(eval UNWIND=--unwind 2)
```
It has always been a slight inconvenience that the starter kit sets `--unwind 1` because this overrides CBMC's ability to figure out on its own how many times to unroll a loop with constant bounds.

template-for-repository/proofs/Makefile.common
```diff
@@ -617,6 +618,11 @@ else
 endif
 
 # Optionally apply loop contracts
+#
+# When applied, the instrumentation introduces a backjump to prove both
+# the base case and then the step case on the same loop.
+# Therefore, it is necessary to unwind this twice: once for the base case,
+# and once more for the step case.
```
markrtuttle, line 625: Could we rephrase this to be less passive:

Unroll loops twice when applying loop contracts.  The instrumentation for loop contracts introduces a back jump to prove both the base case and the induction step of the induction, and we need to unroll both iterations.

markrtuttle (changes requested):

markrtuttle:
I recommend closing this in favor of https://github.com/model-checking/cbmc-starter-kit/pull/110.

PR #91, opened 2022-03-25
Simplify setup scripts

markrtuttle:
This pull request simplifies the setup scripts by adding functions that answer most of the questions about the repository layout that they have been asking the user.   Examples are the location of the repository root, the proof root, and the litani submodule.  Adding the starter kit to a project and configuring for a proof of pvPortMalloc in FreeRTOS is now just

```
cbmc: ./starter-kit/scripts/setup.py 
What is the project name? Kernel
cbmc: cd proofs/

cbmc/proofs: ../starter-kit/scripts/setup-proof.py 
What is the function name? pvPortMalloc
These source files define a function 'pvPortMalloc':
   0 ../../../portable/ARMv8M/secure/heap/secure_heap.c
   1 ../../../portable/GCC/ARM_CM23/secure/secure_heap.c
   2 ../../../portable/GCC/ARM_CM33/secure/secure_heap.c
   3 ../../../portable/IAR/ARM_CM23/secure/secure_heap.c
   4 ../../../portable/IAR/ARM_CM33/secure/secure_heap.c
   5 ../../../portable/MemMang/heap_1.c
   6 ../../../portable/MemMang/heap_2.c
   7 ../../../portable/MemMang/heap_3.c
   8 ../../../portable/MemMang/heap_4.c
   9 ../../../portable/MemMang/heap_5.c
  10 ../../../portable/WizC/PIC18/port.c
  11 The source file is not listed here
Select a source file (the options are 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11): 9
cbmc/proofs: cd pvPortMalloc

cbmc/proofs/pvPortMalloc: make
```

### DIFF
--- a/scripts/repository.py
+++ b/scripts/repository.py
@@ -1,0 +1,229 @@
+"""Discover repository properties like repository root, proof root, etc."""
+
+from pathlib import Path
+import subprocess
+import logging
+import json
+import re
+
+################################################################
+# Construct an ascending relative path like "../../.." from a
+# directory to an ancestor in the file system.
+#
+# The Path method dst.relative_to(src) requires that dst is a
+# descendant of src and will not produce a path like "../../..".
+
+def path_to_ancestor(descendant, ancestor):
+    """Relative path from descendant to ancestor."""
+
+    descendant = Path(descendant).resolve()
+    ancestor = Path(ancestor).resolve()
+
+    try:
+        path = descendant.relative_to(ancestor)
+    except ValueError:
+        raise UserWarning(f"{ancestor} is not an ancestor of {descendant}") from ValueError
+
+    return Path(*[Path('..') for part in path.parts])
+
+################################################################
+# Discover the roots of
+#   * the respository and
+#   * the proofs subtree installed by the starter kit.
+
+def repository_root(cwd='.', abspath=True):
+    """Path to root of repository containing current directory.
+
+    Return the absolute path if abspath is True.  If abspath is False,
+    return the relative path from the current directory.  If the
+    current directory is within a submodule of the repository, then
+    the root of the submodule is returned, not the repository itself.
+    """
+
+    cwd = Path(cwd).resolve()
+    for ancestor in [cwd, *cwd.parents]:
+        if (ancestor / '.git').is_dir():
+            return ancestor if abspath else path_to_ancestor(cwd, ancestor)
+    raise UserWarning(f"No git repository contains {cwd}")
+
+def proofs_root(cwd='.', abspath=True):
+    """Path to root of proofs subtree installed by starter kit.
+
+    Return an absolute path if abspath is True, and a path relative to
+    cwd otherwise.  Search starts at cwd and ascends until it reaches
+    the root of the enclosing repository, and raises UserWarning if
+    there is no enclosing repository."""
+
+    cwd = Path(cwd).resolve()                     # Where to start looking
+    root = repository_root(cwd=cwd, abspath=True) # Where to stop looking
+    proofs = "proofs"                             # What to look for
+
+    for path in [cwd, *cwd.parents]:
+        if path.name == proofs:
+            return path if abspath else path_to_ancestor(cwd, path)
+        if path == root:
+            break
+    raise UserWarning(f"'{cwd}' has no ancestor named '{proofs}'")
+
+################################################################
+# Discover the roots of
+#   * the litani submodule
+#   * the starter kit submodule
+
+def load_submodules(repo=None):
+    """Load file .gitmodules describing the installed submodules"""
+
+    repo = repo or repository_root()
+    try:
+        cmd = ['git', 'config', '-f', '.gitmodules', '--list']
+        kwds = {'cwd': str(repo), 'capture_output': True, 'text': True}
+        lines = subprocess.run(cmd, **kwds, check=True).stdout.splitlines()
+    except subprocess.CalledProcessError as error:
+        logging.debug(error)
+        raise UserWarning(
+            f"Can't load submodules in repository root '{repo}'"
+        ) from error
+
+    submodules = {}
+    for line in lines:
+        line=re.sub(r'\s+', ' ', line.strip())
+
+        # Output of git config is lines of the form
+        #   submodule.ORGANIZATION/REPOSITORY.path=PATH
+        #   submodule.ORGANIZATION/REPOSITORY.url=URL
+        match = re.match(r"^submodule\.([^=]+)\.([^=]+) *= *(.+)", line)
+        if not match:
+            logging.debug("Can't parse git config output: '%s'", line)
+            continue
+
+        name, key, value = [string.strip() for string in match.groups()[0:3]]
+        submodules[name] = submodules.get(name, {})
+        submodules[name][key] = value
+    return submodules
+
+def submodule_root(url, submodules=None, repo=None, abspath=True):
+    """Look up path to root of submodule url in submodules."""
+
+    url = url.lower()
+    repo = repo or repository_root()
+    submodules = submodules or load_submodules(repo=repo)
+
+    for _, config in submodules.items():
+        config_url = config.get("url")
+        config_path = config.get("path")
+
+        if not config_url or not config_url.lower() in [url, url+".git"]:
+            continue
+        if not config_path:
+            logging.debug("Found submodule config = %s", config)
+            raise UserWarning(f"Can't find path to submodule '{url}'")
+        return Path(repo/config_path if abspath else config_path)
+
+    logging.debug("Found submodules = %s", submodules)
+    raise UserWarning(f"Can't find submodule '{url}'")
+
+def litani_root(submodules=None, repo=None, abspath=True):
+    """Root of litani submodule."""
+
+    litani = 'https://github.com/awslabs/aws-build-accumulator'
+    return submodule_root(litani, submodules, repo, abspath)
+
+def starter_kit_root(submodules=None, repo=None, abspath=True):
+    """Root of starter kit submodule."""
+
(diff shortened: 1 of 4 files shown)


scripts/repository.py
```diff
@@ -0,0 +1,229 @@
...
+                srcs.append(path.relative_to(repo))
+            except ValueError: # path is not a file or not relative to repo
+                logging.debug("Skipping %s: not a file under %s", src, repo)
+                continue
+        return [str(src) for src in srcs]
+
+    repo = Path(repo).resolve()
+    sources = sources_under_root(sources, repo)
+    try:
+        cmd = [
```
karkhaz, line 186: Creating a command line this long isn't going to work in general, with a few thousand C files you might easily have a command line longer than ARG_MAX.

Suggest to create a `Popen` object with the options you have below plus `stdin=subprocess.PIPE`, and also pass `"-L", "-"` to `ctags`. Then you can use `communicate()` to send the newline-delimited list of C files to ctags' stdin. Again, the output is probably going to be too big here for `run` anyway.

More generally, is there a reason why you don't just use the ctags `--recurse` option?

scripts/setup-proof.py
```diff
@@ -17,16 +18,16 @@ def proof_template_filenames():
 
 def read_proof_template(filename):
     directory = os.path.join(util.templates_root(), util.PROOF_TEMPLATES)
-    with open(os.path.join(directory, filename)) as data:
+    with open(os.path.join(directory, filename), encoding='utf-8') as data:
```
karkhaz, line 21: just curious, why is this important?

scripts/setup.py
```diff
@@ -52,23 +54,16 @@ def main():
...
-
-    # the script is linking to the litani script within the litani submodule
-    litani = util.read_litani_path()
-
-    # the name of the project used in project verification reports
-    project_name = util.read_project_name()
+    cbmc_root = Path.cwd()
+    proof_root = cbmc_root / "proofs"
+    source_root = repository.repository_root()
+    litani = repository.litani_root() / 'litani'
```
karkhaz, line 60: `submodule_root` concerns me because it raises on failure. In future commits, we plan to use the Litani submodule only if it already is in the repo, and to check for Litani on the PATH otherwise. But that doesn't work if the method throws on failure.

I guess eventually the code will look like this?
```
try:
   litani = repository.litani_root() / "litani"
catch UserWarning:
    litani = os.which("litani") or raise UserWarning("no submodule and no path")
```
?

karkhaz (changes requested):

scripts/repository.py
markrtuttle: I understand, but this method is VASTLY faster.  I don't understand why.

scripts/repository.py
```diff
@@ -0,0 +1,229 @@
...
+        ) from error
+
+    submodules = {}
+    for line in lines:
+        line=re.sub(r'\s+', ' ', line.strip())
+
+        # Output of git config is lines of the form
+        #   submodule.ORGANIZATION/REPOSITORY.path=PATH
+        #   submodule.ORGANIZATION/REPOSITORY.url=URL
+        match = re.match(r"^submodule\.([^=]+)\.([^=]+) *= *(.+)", line)
```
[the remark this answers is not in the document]
markrtuttle: I prefer the regular expressions to be trivial and clean up whitespace below.

scripts/util.py
```diff
@@ -36,34 +38,47 @@ def templates_root():
...
+        for index, source in enumerate(options):
+            print(f" {index:3} {source}")
+        index = input(
+            f"Select a source file (the options are {', '.join(choices)}): "
+        ).strip()
 
-def read_project_name():
-    print("What is the project name? ", end="")
-    return read_from_stdin()
+    if not index in choices:
```
angelonakos, line 71: In the example that you provided in the PR's description, `choices` is equal to: `['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11']`

If a user were to enter `12` then `not index in choices` evaluates to `True` and the UserWarning would be raised (as it should).

Both `index not in choices` and `not index in choices` result into the same value. I tend to prefer the first in terms of readability.

scripts/repository.py
```diff
@@ -0,0 +1,229 @@
...
+    """Root of litani submodule."""
+
+    litani = 'https://github.com/awslabs/aws-build-accumulator'
+    return submodule_root(litani, submodules, repo, abspath)
+
+def starter_kit_root(submodules=None, repo=None, abspath=True):
+    """Root of starter kit submodule."""
+
+    old_starter = 'https://github.com/awslabs/aws-templates-for-cbmc-proofs'
+    new_starter = 'https://github.com/model-checking/cbmc-starter-kit'
```
angelonakos, line 135: Should we also be trying the following 2?

```
git@github.com:awslabs/aws-templates-for-cbmc-proofs.git
git@github.com:model-checking/cbmc-starter-kit.git
```

scripts/repository.py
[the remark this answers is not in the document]
markrtuttle: Simplified to "^submodule\.(.+)\.(path|url)=(.+)".  Parsing with regular expressions will be unreliable if PATH or URL can contain path= or url= as substrings.  So use a simple regular expression here and confirm that PATH is an existing directory below.

scripts/repository.py
```diff
@@ -0,0 +1,229 @@
...
+
+        # Output of git config is lines of the form
+        #   submodule.ORGANIZATION/REPOSITORY.path=PATH
+        #   submodule.ORGANIZATION/REPOSITORY.url=URL
+        match = re.match(r"^submodule\.([^=]+)\.([^=]+) *= *(.+)", line)
+        if not match:
+            logging.debug("Can't parse git config output: '%s'", line)
+            continue
+
+        name, key, value = [string.strip() for string in match.groups()[0:3]]
```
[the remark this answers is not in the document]
markrtuttle: Dissonance made consonant.

scripts/repository.py
```diff
@@ -0,0 +1,229 @@
...
+    for _, config in submodules.items():
+        config_url = config.get("url")
+        config_path = config.get("path")
+
+        if not config_url or not config_url.lower() in [url, url+".git"]:
+            continue
+        if not config_path:
+            logging.debug("Found submodule config = %s", config)
+            raise UserWarning(f"Can't find path to submodule '{url}'")
+        return Path(repo/config_path if abspath else config_path)
```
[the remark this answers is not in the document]
markrtuttle: Submodule parsing used strings for path and url, so config_path was a string, but I've fixed parsing to use Path object for path and removed Path altogether here.

scripts/repository.py
```diff
@@ -0,0 +1,229 @@
...
+
+    litani = 'https://github.com/awslabs/aws-build-accumulator'
+    return submodule_root(litani, submodules, repo, abspath)
+
+def starter_kit_root(submodules=None, repo=None, abspath=True):
+    """Root of starter kit submodule."""
+
+    old_starter = 'https://github.com/awslabs/aws-templates-for-cbmc-proofs'
+    new_starter = 'https://github.com/model-checking/cbmc-starter-kit'
+    return (submodule_root(old_starter, submodules, repo, abspath) or
```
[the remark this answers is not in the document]
markrtuttle: Good catch.  Changed submodule_root to return None.  But I require some coaching on when to raise and when to return useless values that have to be checked for at the call site.

scripts/repository.py
markrtuttle: Good catch.

scripts/setup-proof.py
markrtuttle: Important?  Why, because pylint says it is important.  See `pylint --help-msg=W1514`

scripts/setup.py
markrtuttle: I've changed `submodule_root` to return None in light of your prior observations.

scripts/util.py
markrtuttle: Good catch.  I've changed both to 'not in'.

scripts/repository.py
```diff
@@ -0,0 +1,229 @@
...
+    """Paths to all source files within a repository.
+
+    Return a list of paths to C source files found under the
+    repository root (paths are relative to the repository root)."""
+
+    repo = Path(repo).resolve()
+    try:
+        cmd = ['find', '.', '-name', '*.c']
+        kwds = {'cwd': repo, 'capture_output': True, 'text': True}
+        return subprocess.run(cmd, **kwds, check=True).stdout.splitlines()
```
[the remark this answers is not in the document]
markrtuttle: Done

scripts/repository.py
```diff
@@ -0,0 +1,229 @@
...
+        srcs = []
+        for src in sources:
+            path = (repo / src).resolve()
+            try:
+                if not path.is_file():
+                    raise ValueError
+                srcs.append(path.relative_to(repo))
+            except ValueError: # path is not a file or not relative to repo
+                logging.debug("Skipping %s: not a file under %s", src, repo)
+                continue
```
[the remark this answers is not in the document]
markrtuttle: Deleted this function.

scripts/repository.py
markrtuttle: Done.  TL;DR: I was using the ctags option --filter=yes and should have been using -L - (the --help documentation made them sound the same).

karkhaz (approved):

angelonakos (approved):

markrtuttle:
@angelonakos That is a nice change.  I've amplified on it an improved handling of the default choice for `ask_for_source_file`, see the [new definition](https://github.com/model-checking/cbmc-starter-kit/blob/7ca9b3a3783a280e13e6923d830a7d2559a3aa6a/scripts/util.py#L53) and search for instances of the variable `index`.

PR #92, opened 2022-03-31
Add support for remove-static via crangler

tautschnig:
crangler is a utility shipped with CBMC that permits (among other
features) selective removal of "static" from functions and objects via
JSON configuration files. From CBMC's perspective this is the preferred
route forward instead of using --export-file-local-symbols. Key
advantages are that selective removal is possible, and that there is no
renaming of functions. The latter makes for more intuitive proof
harnesses.

### DIFF
--- a/src/cbmc_starter_kit/template-for-proof/Makefile
+++ b/src/cbmc_starter_kit/template-for-proof/Makefile
@@ -23,13 +23,13 @@ PROJECT_SOURCES += $(SRCDIR)/<__PATH_TO_SOURCE_FILE__>
 # If you require access to a file-local ("static") function or object to conduct
 # your proof, set the following (and do not include the original source file
 # ("<__PATH_TO_SOURCE_FILE__>") in PROJECT_SOURCES).
-# REMOVE_STATIC = $(PROOFDIR)/<__SOURCE_FILE_BASENAME__>.i
+# REWRITTEN_SOURCES = $(PROOFDIR)/<__SOURCE_FILE_BASENAME__>.i
 # include <__PATH_TO_MAKEFILE__>/Makefile.common
 # $(PROOFDIR)/<__SOURCE_FILE_BASENAME__>.i_SOURCE = $(SRCDIR)/<__PATH_TO_SOURCE_FILE__>
 # $(PROOFDIR)/<__SOURCE_FILE_BASENAME__>.i_FUNCTIONS = foo bar
 # $(PROOFDIR)/<__SOURCE_FILE_BASENAME__>.i_OBJECTS = baz
-# Care is required with variables on the left-hand side: REMOVE_STATIC must be
-# set before including Makefile.common, but any use of variables on the
+# Care is required with variables on the left-hand side: REWRITTEN_SOURCES must
+# be set before including Makefile.common, but any use of variables on the
 # left-hand side requires those variables to be defined. Hence, _SOURCE,
 # _FUNCTIONS, _OBJECTS is set after including Makefile.common.
 
--- a/src/cbmc_starter_kit/template-for-proof/Makefile
+++ b/src/cbmc_starter_kit/template-for-proof/Makefile
@@ -20,4 +20,17 @@ PROJECT_SOURCES += $(SRCDIR)/<__PATH_TO_SOURCE_FILE__>
 # documentation in Makefile.common under the "Job Pools" heading for details.
 # EXPENSIVE = true
 
+# If you require access to a file-local ("static") function or object to conduct
+# your proof, set the following (and do not include the original source file
+# ("<__PATH_TO_SOURCE_FILE__>") in PROJECT_SOURCES).
+# REMOVE_STATIC = $(PROOFDIR)/<__SOURCE_FILE_BASENAME__>.i
+# include <__PATH_TO_MAKEFILE__>/Makefile.common
+# $(PROOFDIR)/<__SOURCE_FILE_BASENAME__>.i_SOURCE = $(SRCDIR)/<__PATH_TO_SOURCE_FILE__>
+# $(PROOFDIR)/<__SOURCE_FILE_BASENAME__>.i_FUNCTIONS = foo bar
+# $(PROOFDIR)/<__SOURCE_FILE_BASENAME__>.i_OBJECTS = baz
+# Care is required with variables on the left-hand side: REMOVE_STATIC must be
+# set before including Makefile.common, but any use of variables on the
+# left-hand side requires those variables to be defined. Hence, _SOURCE,
+# _FUNCTIONS, _OBJECTS is set after including Makefile.common.
+
 include <__PATH_TO_MAKEFILE__>/Makefile.common
--- a/src/cbmc_starter_kit/template-for-repository/proofs/Makefile-project-defines
+++ b/src/cbmc_starter_kit/template-for-repository/proofs/Makefile-project-defines
@@ -16,6 +16,11 @@
 # Flags to pass to goto-cc for linking (typically those passed to gcc)
 # LINK_FLAGS =
 
+# Flag to pass to goto-cc to make all static symbols globally visible. Leave it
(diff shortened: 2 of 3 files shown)


danielsn:
What happens if two files have static functions with different signatures but the same name?

tautschnig:
> What happens if two files have static functions with different signatures but the same name?

This will result in a linking warning. Adding support for renaming in `crangler` is possible, but I'd prefer to wait until an actual usecase becomes available so that we can have decent testing in place.

tautschnig:
> Please demonstrate this work on the proof [findHeaderFieldParserCallback](https://github.com/FreeRTOS/coreHTTP/tree/main/test/cbmc/proofs/findHeaderFieldParserCallback).
> 

Here is the diff:

````patch
diff --git a/test/cbmc/proofs/findHeaderFieldParserCallback/Makefile b/test/cbmc/proofs/findHeaderFieldParserCallback/Makefile
index cb66ad2..4c445d1 100644
--- a/test/cbmc/proofs/findHeaderFieldParserCallback/Makefile
+++ b/test/cbmc/proofs/findHeaderFieldParserCallback/Makefile
@@ -13,11 +13,13 @@ DEFINES += -DMAX_HEADER_FIELD_LENGTH=$(MAX_HEADER_FIELD_LENGTH)
 INCLUDES +=

 REMOVE_FUNCTION_BODY +=
-UNWINDSET += __CPROVER_file_local_core_http_client_c_caseInsensitiveStringCmp.0:$(MAX_HEADER_FIELD_LENGTH)
+UNWINDSET += caseInsensitiveStringCmp.0:$(MAX_HEADER_FIELD_LENGTH)

 PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
 PROOF_SOURCES += $(SRCDIR)/test/cbmc/sources/http_cbmc_state.c

-PROJECT_SOURCES += $(SRCDIR)/source/core_http_client.c
-
+EXPORT_FILE_LOCAL_SYMBOLS =
+REMOVE_STATIC = $(PROOFDIR)/core_http_client.i
 include ../Makefile.common
+$(PROOFDIR)/core_http_client.i_SOURCE = $(SRCDIR)/source/core_http_client.c
+$(PROOFDIR)/core_http_client.i_FUNCTIONS = findHeaderFieldParserCallback
diff --git a/test/cbmc/proofs/findHeaderFieldParserCallback/findHeaderFieldParserCallback_harness.c b/test/cbmc/proofs/findHeaderFieldParserCallback/findHeaderFieldParserCallback_harness.c
index 8c5399b..94dea9c 100644
--- a/test/cbmc/proofs/findHeaderFieldParserCallback/findHeaderFieldParserCallback_harness.c
+++ b/test/cbmc/proofs/findHeaderFieldParserCallback/findHeaderFieldParserCallback_harness.c
@@ -29,9 +29,9 @@
 #include "llhttp.h"
 #include "core_http_client.h"

-int __CPROVER_file_local_core_http_client_c_findHeaderFieldParserCallback( llhttp_t * pHttpParser,
-                                                                           const char * pFieldLoc,
-                                                                           size_t fieldLen );
+int findHeaderFieldParserCallback( llhttp_t * pHttpParser,
+                                   const char * pFieldLoc,
+                                   size_t fieldLen );


 void findHeaderFieldParserCallback_harness()
@@ -62,5 +62,5 @@ void findHeaderFieldParserCallback_harness()
     pFindHeaderContext->valueFound = 0;
     pHttpParser->data = ( void * ) pFindHeaderContext;

-    __CPROVER_file_local_core_http_client_c_findHeaderFieldParserCallback( pHttpParser, pFieldLoc, fieldLen );
+    findHeaderFieldParserCallback( pHttpParser, pFieldLoc, fieldLen );
 }
````
resulting in [this report](https://github.com/model-checking/cbmc-starter-kit/files/8602451/report.zip).

> Consider merging this pull request with #93 and using definitions of EXPORT_FILE_LOCAL_SYMBOLS in Makefile.common and Makefile-project-defines.

Done.

**Comment by remi-delmas-3000** [comment]
The example shows
``` 
+EXPORT_FILE_LOCAL_SYMBOLS =
+REMOVE_STATIC = $(PROOFDIR)/core_http_client.i
```
Can we document what users are supposed to do to put these variables in a consistent state ? Or define one from the other in the `Makefile.common` to make them consistent ?

**Comment by tautschnig** [comment]
> Can we document what users are supposed to do to put these variables in a consistent state ? Or define one from the other in the `Makefile.common` to make them consistent ?

1) The need for `EXPORT_FILE_LOCAL_SYMBOLS =` only arises as this project otherwise relies on all file-local symbols being exported. New projects will not require this.
2) It's actually ok if `EXPORT_FILE_LOCAL_SYMBOLS =` is not set, it will just mean that not only the selected function(s) are made non-static, but all other static functions will be as well and those others will be renamed.

**Comment by jimgrundy** [comment]
Which functions are being made non-static in this case?
- Is it all static function in the selected file? That seems like a blunt instrument. Can it be done selectively for a specific function? Can we demonstrate it being done selectively?
- What happens if making a function non-static results in a name-clash? I had some sympathy for the __CBMC_file_local_filename_mumble approach, verbose as it was, because it greatly reduced the chance of that.
- Can we have a mechanism where we say "make function foo in file bar nonstatic and call the resulting function bar_foo", thereby empowering the user to pick a name to avoid any clash?
- Can we also do this for static variables as well as functions (we need to)?

**Comment by jimgrundy** [comment]
OK, having reviewed with Michael, I'm OK with this PR.

**Comment by nwetzler** [comment]
> Which functions are being made non-static in this case?
> 
> * Is it all static function in the selected file? That seems like a blunt instrument. Can it be done selectively for a specific function? Can we demonstrate it being done selectively?
> * What happens if making a function non-static results in a name-clash? I had some sympathy for the __CBMC_file_local_filename_mumble approach, verbose as it was, because it greatly reduced the chance of that.
> * Can we have a mechanism where we say "make function foo in file bar nonstatic and call the resulting function bar_foo", thereby empowering the user to pick a name to avoid any clash?
> * Can we also do this for static variables as well as functions (we need to)?

It seems like you and Jim synced, but I'd be interested in the answers to these out of curiosity.

**Comment by nwetzler** [comment]
I've created a branch on s2n-tls testing out the crangler integration with the starter kit.
https://github.com/nwetzler/private-s2n-cbmc/tree/crangler-test

I'm running into two separate problems.  The first is an invariant failure when using crangler to initialize `s2n_mem_free_cb` in the `s2n_connection_free_managed_send_io` harness.
```
Enforcing contracts
--- begin invariant violation report ---
Invariant check failed
/tmp/cbmc-20220609-56930-1dxvwl4/src/goto-programs/goto_program.h:650 function: insert_before_swap
Condition: target!=instructions.end()
Reason: Precondition
```

The second problem is when using multiple REMOVE_STATIC declarations in the Makefile resulting in the following error message:
`../Makefile.common:409: *** target pattern contains no `%'.  Stop.`

tautschnig:
Thanks @nwetzler for trying this out! I'll try to debug both of these issues over the next hours.

tautschnig:
@nwetzler I have now also diagnosed the second issue. We should likely add a more useful error message, but the root cause is that you are trying to use `--enforce-contract` with a function that does not have a function body (anymore). I'm saying not "anymore," because the function is `__CPROVER_file_local_s2n_connection_c_s2n_connection_free_managed_send_io` when it should now be `s2n_connection_free_managed_send_io`.

nwetzler:
> I have updated this branch again to address this issue (I previously hadn't picked up the defines, implying that `crangler` was preprocessing without `-DCBMC=1`). That said, I also noticed that a fix in `crangler` may be necessary for some assigns clauses, but I'm not sure this is a case that occurs in s2n.

I now have a couple of our s2n examples with function contracts working and have updated my branch:
https://github.com/nwetzler/private-s2n-cbmc/tree/crangler-test

markrtuttle (approved):

nwetzler:
The change to documentation is nice.  Cheers.

PR #93, opened 2022-03-31
Do not unconditionally compile with --export-file-local-symbols

tautschnig:
Users should have a choice to not exporting all static symbols by
suitably setting COMPILE_FLAGS themselves.


tautschnig:
Closing as it will be made part of #92.

PR #104, opened 2022-04-18
Run starter-kit tests in CI

ronakfof:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -1,0 +1,39 @@
+name: Run Tests
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+  push:
+    branches: [ master ]
+
+jobs:
+  run-tests:
+    if: "!contains(github.event.pull_request.labels.*.name, 'no-test')"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yaml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          curl -o litani.deb -L https://github.com/awslabs/aws-build-accumulator/releases/download/1.24.0/litani-1.24.0.deb
+          curl -o cbmc.deb -L https://github.com/diffblue/cbmc/releases/download/cbmc-5.54.0/ubuntu-20.04-cbmc-5.54.0-Linux.deb
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+                                  ./litani.deb \
+                                  ./cbmc.deb \
+                                  universal-ctags
+          rm -rf litani.deb cbmc.deb
+          python3 -m pip install cbmc-viewer cbmc-starter-kit
```
markrtuttle, line 32: Is this right?  Don't we want to test the current commit and not the last published starter kit?

ronakfof: I just created a boilerplate test. Was hoping you'll comment with the test for both `update` and `setup` and I can then update the workflow.

angelonakos:
Closing in favor of #122

PR #106, opened 2022-04-22
Add bash completion script

markrtuttle:
Adds completion for starter kit command line options.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,6 +32,8 @@ where = src
 # We want to two subtrees of data, but static configuration files do
 # not appear to give us an easy way to do this.
 cbmc_starter_kit =
+  etc/*
+  etc/bash_completion.d/*
   template-for-proof/*
   template-for-proof/*/*
   template-for-proof/*/*/*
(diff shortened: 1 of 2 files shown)


src/cbmc_starter_kit/etc/cbmc-starter-kit-bash-completion.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
```
tautschnig, line 1: Please install in `etc/bash_completion.d/` for it to be automatically found on most systems when bash completion is enabled.

tautschnig (approved):

src/cbmc_starter_kit/etc/cbmc-starter-kit-bash-completion.sh
angelonakos:
```bash
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

could be added

angelonakos (approved):

src/cbmc_starter_kit/etc/cbmc-starter-kit-bash-completion.sh
markrtuttle: Added MIT-0 license to all source files.  Moved bash completion to etc/bash_completion.d/cbmc-starter-kit.sh, but as discussed earlier, we can install under brew but there is no os-independent way to install under pip.

Issue #116, opened 2022-05-19
Prior to execution, the Python "run script" should delete Litani cache pointers present inside individual proof directories


angelonakos:
** Issue **

If the directory for a particular CBMC proof has the .litani_cache_dir file present, then the Python "run script" will fail.

** Acceptance Criteria **

- If one or multiple CBMC proof directories have "Litani cache pointers" present, then no failure will be encountered when running the Python "run script" from the directory containing all proofs.

tautschnig:
Resolving as #117 has been merged.

PR #117, opened 2022-05-19
Remove .litani_cache_dir from individual directory before considering proof

angelonakos:
*Issue #, if available:*

#116 

*Description of changes:*

If `make report` is executed within an individual CBMC proof directory, Litani
will create its "cache pointer" (a file called .litani_cache_dir).

This change makes sure that such a file is deleted prior to that individual CBMC
proof directory being included in the list of all proof directories.

### DIFF
--- a/src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
+++ b/src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
@@ -180,13 +180,13 @@ def get_proof_dirs(proof_root, proof_list, marker_file):
 
     for root, _, fyles in os.walk(proof_root):
         proof_name = str(pathlib.Path(root).name)
+        if root != str(proof_root) and ".litani_cache_dir" in fyles:
+            pathlib.Path(f"{root}/.litani_cache_dir").unlink()
         if proof_list and proof_name not in proof_list:
             continue
         if proof_list and proof_name in proofs_remaining:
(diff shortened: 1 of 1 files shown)


src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -185,6 +185,8 @@ def get_proof_dirs(proof_root, proof_list, marker_file):
         if proof_list and proof_name in proofs_remaining:
             proofs_remaining.remove(proof_name)
         if marker_file in fyles:
+            if ".litani_cache_dir" in fyles:
+                pathlib.Path(f"{proof_name}/.litani_cache_dir").unlink()
```
markrtuttle, line 189: I would use the python equivalent of "rm -r $(find . -name .litani_cache_dir)" to cover the unlikely case that the the proofs are not a flat collection of proof directories under "proofs" and there is a stray litani cache directory in some intermediate directory.

markrtuttle (approved):

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
angelonakos: Incorporated your suggestion. Now, the Litani artifact will be deleted from every directory, except for the PROOFS_DIR itself.

PR #118, opened 2022-05-19
Enable debug output in run-cbmc-proofs.py script

angelonakos:
*Issue #, if available:*

*Description of changes:*

This PR add a `--debug` flag to the "run script", such that
both "stdout" and "stderr" of the "make _report" are
logged.

### DIFF
--- a/src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
+++ b/src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
@@ -140,6 +140,10 @@ def get_args():
             "action": "store_true",
             "help": "verbose output",
     }, {
+            "flags": ["--debug"],
+            "action": "store_true",
+            "help": "debug output",
+    }, {
             "flags": ["--version"],
             "action": "version",
             "version": "_CBMC_STARTER_KIT_VERSION_",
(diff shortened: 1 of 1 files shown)


src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -287,11 +291,25 @@ async def configure_proof_dirs(
...
+                logging.debug(line)
+            logging.debug("stderr:")
+            for line in stderr.decode().splitlines():
+                logging.debug(line)
+        else:
+            proc = await asyncio.create_subprocess_exec(
+                "nice", "-n", "15", "make", *pools, *profiling, "-B", "--quiet",
+                "_report", cwd=path)
+            await proc.wait()
+
```
markrtuttle, line 312: Don't you want to use a single block of code (not if-then-else) and just use 
```
"VERBOSE=1" if debug else ""
```
The debug output will be generated only when the logging level is set to debug.

markrtuttle (changes requested):

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
angelonakos: > The debug output will be generated only when the logging level is set to debug.

That is right. The if-then-else block was indeed unnecessary.

markrtuttle (approved):

PR #119, opened 2022-05-20
Support multiple variants of ctags

markrtuttle:
This pull request copies from cbmc-viewer the support for all major variants of ctags.

Also add support for users specifying source files using pathnames including ~ or ~user and using relative pathnames.

Also advance version to 2.3.

### DIFF
--- a/src/cbmc_starter_kit/ctagst.py
+++ b/src/cbmc_starter_kit/ctagst.py
@@ -1,0 +1,158 @@
+"""Ctags support for locating symbol definitions"""
+
+from pathlib import Path
+import json
+import logging
+import subprocess
+import sys
+
+################################################################
+# This popen method is used to subprocess-out the invocation of ctags.
+# This method duplicates code in other modules to make this ctags
+# module a stand-alone module that can be copied and reused in other
+# projects.
+
+def popen(cmd, cwd=None, stdin=None, encoding=None):
+    """Run a command with string stdin on stdin, return stdout and stderr."""
+
+    cmd = [str(word) for word in cmd]
+    kwds = {'cwd': cwd,
+            'text': True,
+            'stdin': subprocess.PIPE,
+            'stdout': subprocess.PIPE,
+            'stderr': subprocess.PIPE}
+    if sys.version_info >= (3, 6): # encoding is new in Python 3.6
+        kwds['encoding'] = encoding or 'utf-8'
+    try:
+        logging.debug('Popen command: "%s"', ' '.join(cmd))
+        logging.debug('Popen stdin: "%s"', stdin)
+        with subprocess.Popen(cmd, **kwds) as pipe:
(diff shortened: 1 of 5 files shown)


angelonakos (approved):

src/cbmc_starter_kit/ctagst.py
```diff
@@ -0,0 +1,158 @@
...
+        '-f', '-', # write tags to standard output, one tag per line
+        '--output-format=json', # each tag is a one-line json blob
+        '--fields=FNnK' # json blob is {"name": symbol, "path": file, "line": line, "kind": kind}
+    ]
+    try:
+        logging.info("Running universal ctags")
+        stdout, _ = popen(cmd, cwd=root, stdin='\n'.join(files))
+        strings = stdout.splitlines()
+    except UserWarning:
+        logging.info("Universal ctags failed")
```
karkhaz, line 72: Suggest print out stderr when this happens.

With this implementation of popen, users currently need to rerun with `--debug` to get the actual error message from ctags. (and they need to know to run with `--debug`).

src/cbmc_starter_kit/ctagst.py
```diff
@@ -0,0 +1,158 @@
...
+################################################################
+
+def legacy_ctags(root, files):
+    """Use legacy ctags to list symbols defined in files under root."""
+
+    # MacOS ships with a legacy ctags from BSD installed in /usr/bin/ctags.
+    # See the MacOS man page for the documentation used to implement this method.
+    cmd = ['ctags',
+           '-x',  # write human-readable summary to standard output
+           *files # legacy ctags cannot read list of files from stdin
```
karkhaz, line 136: Is the list of files ever expected to be large? The operating system has a maximum length for argument lists, and it varies between OSs. I guess the catch will deal with this failing to launch though.
```
# on macOS
; getconf ARG_MAX
1048576
```

karkhaz (approved):

src/cbmc_starter_kit/ctagst.py
markrtuttle: No, the ctags() method invokes ctags as universal ctags, then exhuberant ctags, then legacy ctags, and uses the result of the first successful invocation (if any).   I don't want error messages printed during the fail down.  You might ask me to check the version and invoke the right version, but universal ctags can be built in many configurations and if the nice invocation of universal fails because it is missing the json extension, for example, I want to invoke using the exhuberant arguments that are reverse compatible.

src/cbmc_starter_kit/ctagst.py
markrtuttle: The invocation will fail, and the only symbols that will be linked are the symbols that actually appear in the cbmc symbol table.  That's not so awful.  And the right solution is to install a more modern version of ctags (the legacy is from BSD 3.0 and earlier).

PR #120, opened 2022-05-24
Summarize execution of CBMC proofs by printing a JSON object to stdout

angelonakos:
*Issue #, if available:*

*Description of changes:*

A user will be able to get a breakdown of statuses across
"proof pipelines" as well as a breakdown of the status of
each "proof pipeline",

This PR is dependent on https://github.com/awslabs/aws-build-accumulator/pull/167


karkhaz:
This is for users running proofs locally. In my opinion, diagnostic information printed at the end should be human-readable. If users want a JSON blob, they can use litani's actual output, or it should be a separate flag that dumps JSON.

angelonakos:
Cancelling in favor of #121

PR #121, opened 2022-05-25
    Add optional flag to Python run-script that summarizes CBMC proofs results

angelonakos:
*Issue #, if available:*

*Description of changes:*

The Python run-script will now print 2 tables in GitHub flavored Markdown, which summarize CBMC proof results. If the `--summarize` flag is used then these tables will be printed.

### DIFF
--- a/src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
+++ b/src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
@@ -1,0 +1,92 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+
+import json
+import logging
+
+
+def _get_max_length_per_column_list(data):
+    ret = [len(item) + 1 for item in data[0]]
+    for row in data[1:]:
+        for idx, item in enumerate(row):
+            ret[idx] = max(ret[idx], len(item) + 1)
+    return ret
+
+
+def _get_table_header_separator(max_length_per_column_list):
+    line_sep = ""
+    for max_length_of_word_in_col in max_length_per_column_list:
+        line_sep += "|" + "-" * (max_length_of_word_in_col + 1)
+    line_sep += "|\n"
+    return line_sep
+
+
+def _get_entries(max_length_per_column_list, row_data):
+    entries = []
+    for row in row_data:
+        entry = ""
+        for idx, word in enumerate(row):
+            max_length_of_word_in_col = max_length_per_column_list[idx]
+            space_formatted_word = (max_length_of_word_in_col - len(word)) * " "
+            entry += "| " + word + space_formatted_word
+        entry += "|\n"
+        entries.append(entry)
+    return entries
+
+
+def _get_rendered_table(data):
+    table = []
+    max_length_per_column_list = _get_max_length_per_column_list(data)
+    entries = _get_entries(max_length_per_column_list, data)
+    for idx, entry in enumerate(entries):
+        if idx == 1:
+            line_sep = _get_table_header_separator(max_length_per_column_list)
+            table.append(line_sep)
+        table.append(entry)
+    table.append("\n")
+    return "".join(table)
+
+
+def _get_status_and_proof_summaries(run_dict):
+    """Parse a dict representing a Litani run and create lists summarizing the
+    proof results.
+
+    Parameters
+    ----------
+    run_dict
+        A dictionary representing a Litani run.
+
+    Returns
+    -------
+    A list of 2 lists.
+    The first sub-list maps a status to the number of proofs with that status.
+    The second sub-list maps each proof to its status.
+    """
+    count_statuses = {}
+    proofs = [["Proof", "Status"]]
+    for proof_pipeline in run_dict["pipelines"]:
+        status_pretty_name = proof_pipeline["status"].title().replace("_", " ")
+        try:
+            count_statuses[status_pretty_name] += 1
+        except KeyError:
+            count_statuses[status_pretty_name] = 1
+        proof = proof_pipeline["name"]
+        proofs.append([proof, status_pretty_name])
+    statuses = [["Status", "Count"]]
+    for status, count in count_statuses.items():
+        statuses.append([status, str(count)])
+    return [statuses, proofs]
+
+
+def print_proof_results(out_file):
+    """
+    Print 2 strings that summarize the proof results.
+    When printing, each string will render as a GitHub flavored Markdown table.
+    """
+    try:
+        with open(out_file, encoding='utf-8') as run_json:
+            run_dict = json.load(run_json)
+            for summary in _get_status_and_proof_summaries(run_dict):
+                print(_get_rendered_table(summary))
+    except Exception as ex: # pylint: disable=broad-except
+        logging.critical("Could not print results. Exception: %s", str(ex))
--- a/src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
+++ b/src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
@@ -14,6 +14,9 @@ import pathlib
 import re
 import subprocess
 import sys
+import tempfile
+
+from lib.summarize import print_proof_results
 
 
 DESCRIPTION = "Configure and run all CBMC proofs in parallel"
@@ -144,10 +147,14 @@ def get_args():
             "action": "store_true",
             "help": "debug output",
     }, {
+            "flags": ["--summarize"],
+            "action": "store_true",
+            "help": "summarize proof results with two tables on stdout",
+    }, {
             "flags": ["--version"],
             "action": "version",
             "version": "_CBMC_STARTER_KIT_VERSION_",
-            "help": "Display version and exit"
+            "help": "display version and exit"
     }]:
         flags = arg.pop("flags")
         pars.add_argument(*flags, **arg)
@@ -200,12 +207,15 @@ def get_proof_dirs(proof_root, proof_list, marker_file):
         sys.exit(1)
 
 
-def run_build(litani, jobs, fail_on_proof_failure):
+def run_build(litani, jobs, fail_on_proof_failure, summarize):
     cmd = [str(litani), "run-build"]
     if jobs:
         cmd.extend(["-j", str(jobs)])
     if fail_on_proof_failure:
         cmd.append("--fail-on-pipeline-failure")
+    if summarize:
+        out_file = pathlib.Path(tempfile.gettempdir(), "run.json").resolve()
+        cmd.extend(["--out-file", str(out_file)])
 
     logging.debug(" ".join(cmd))
     proc = subprocess.run(cmd, check=False)
@@ -215,6 +225,10 @@ def run_build(litani, jobs, fail_on_proof_failure):
             sys.exit(10)
         logging.critical("Failed to run litani run-build")
         sys.exit(1)
+
+    if summarize:
+        print_proof_results(out_file)
+        out_file.unlink()
 
 
 def get_litani_path(proof_root):
@@ -281,7 +295,7 @@ def should_enable_pools(litani_caps, args):
     return "pools" in litani_caps
 
 
-async def configure_proof_dirs(
+async def configure_proof_dirs( # pylint: disable=too-many-arguments
     queue, counter, proof_uids, enable_pools, enable_memory_profiling, debug):
     while True:
         print_counter(counter)
@@ -392,7 +406,7 @@ async def main(): # pylint: disable=too-many-locals
         sys.exit(1)
 
     if not args.no_standalone:
-        run_build(litani, args.parallel_jobs, args.fail_on_proof_failure)
+        run_build(litani, args.parallel_jobs, args.fail_on_proof_failure, args.summarize)
 
 
 if __name__ == "__main__":
--- a/test/repo/Makefile
+++ b/test/repo/Makefile
@@ -59,11 +59,10 @@ update-starter:
 
 run:
 	$(RM) -r $(REPORT_ROOT)
-	cd $(REPO_ROOT)/$(REPO_PROOF_ROOT) && \
-	  $(LITANI) init --project-name "Report" && \
-	  ./run-cbmc-proofs.py --no-standalone $(REPO_PROOFS) && \
-	  $(CONFIG) --litani-bindir $(LITANI_ROOT) --report-root $(REPORT_ROOT) && \
-	  $(LITANI) run-build
+	cd $(REPO_ROOT)/$(REPO_PROOF_ROOT) && $(LITANI) init --project-name "Report"
+	cd $(REPO_ROOT)/$(REPO_PROOF_ROOT) && ./run-cbmc-proofs.py --no-standalone $(REPO_PROOFS)
+	cd $(REPO_ROOT)/$(REPO_PROOF_ROOT) && $(CONFIG) --litani-bindir $(LITANI_ROOT) --report-root $(REPORT_ROOT)
+	cd $(REPO_ROOT)/$(REPO_PROOF_ROOT) && $(LITANI) run-build
 
 run1:
 	$(MAKE) REPORT_ROOT=$(REPORT_ROOT1) run
--- a/test/summarize_test/sample_run.json
+++ b/test/summarize_test/sample_run.json
@@ -1,0 +1,188 @@
+{
+    "aux": {},
+    "project": "example-build",
+    "version": "1.26.0-rc",
+    "version_major": 1,
+    "version_minor": 26,
+    "version_patch": 0,
+    "release_candidate": true,
+    "run_id": "b018ce5a-e34c-45ee-80bc-209c60622e1f",
+    "stages": [
+      "build",
+      "test",
+      "report"
+    ],
+    "start_time": "2022-05-24T02:48:34Z",
+    "status": "fail",
+    "pools": {},
+    "parallelism": {
+      "trace": [
+        {
+          "running": 1,
+          "finished": 1,
+          "total": 1,
+          "time": "2022-05-24T02:51:28.057125Z"
+        }
+      ],
+      "max_parallelism": 1,
+      "n_proc": 12
+    },
+    "latest_symlink": "/var/folders/bd/w7pff05n5k35735rjvph7dkh0000gs/T/litani/runs/latest",
+    "end_time": "2022-05-24T02:51:28Z",
+    "pipelines": [
+      {
+        "name": "pipe-will-fail",
+        "ci_stages": [
+          {
+            "jobs": [],
+            "progress": 0,
+            "complete": true,
+            "status": "success",
+            "url": "artifacts/pipe-will-fail/build",
+            "name": "build"
+          },
+          {
+            "jobs": [],
+            "progress": 0,
+            "complete": true,
+            "status": "success",
+            "url": "artifacts/pipe-will-fail/test",
+            "name": "test"
+          },
+          {
+            "jobs": [
+              {
+                "wrapper_arguments": {
+                  "subcommand": "exec",
+                  "verbose": false,
+                  "very_verbose": false,
+                  "inputs": null,
+                  "command": "/usr/bin/false",
+                  "outputs": null,
+                  "pipeline_name": "pipe-will-fail",
+                  "ci_stage": "report",
+                  "cwd": null,
+                  "timeout": null,
+                  "timeout_ok": false,
+                  "timeout_ignore": false,
+                  "ignore_returns": null,
+                  "ok_returns": null,
+                  "outcome_table": null,
+                  "interleave_stdout_stderr": false,
+                  "stdout_file": null,
+                  "stderr_file": null,
+                  "pool": null,
+                  "description": null,
+                  "profile_memory": false,
+                  "profile_memory_interval": 10,
+                  "phony_outputs": null,
+                  "tags": null,
+                  "status_file": "/var/folders/bd/w7pff05n5k35735rjvph7dkh0000gs/T/litani/runs/b018ce5a-e34c-45ee-80bc-209c60622e1f/status/0f65a89f-d82a-4a37-8ac6-30172cb8de8b.json",
+                  "job_id": "0f65a89f-d82a-4a37-8ac6-30172cb8de8b"
+                },
+                "complete": true,
+                "start_time": "2022-05-24T02:51:28Z",
+                "timeout_reached": false,
+                "command_return_code": 1,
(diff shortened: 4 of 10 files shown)


angelonakos:
I added a target to the test Makefile, which will run the CBMC proofs and specify `--tabular-summary-print-fmt github`.

As can be seen in the logs of either one of the 3 GitHub Actions status checks, when this option is specified then tables such as the following will be printed to stdout:


| Status   |   Count |
|----------|---------|
| Success  |      13 |

|    | Proof                               | Status   |
|----|-------------------------------------|----------|
|  0 | httpParserOnHeaderValueCallback     | Success  |
|  1 | findHeaderValueParserCallback       | Success  |
|  2 | httpParserOnMessageBeginCallback    | Success  |
|  3 | httpParserOnHeaderFieldCallback     | Success  |
|  4 | findHeaderOnHeaderCompleteCallback  | Success  |
|  5 | httpParserOnStatusCallback          | Success  |
|  6 | httpParserOnHeadersCompleteCallback | Success  |
|  7 | HTTPClient_ReadHeader               | Success  |
|  8 | httpParserOnMessageCompleteCallback | Success  |
|  9 | HTTPClient_AddRangeHeader           | Success  |
| 10 | httpParserOnBodyCallback            | Success  |
| 11 | findHeaderFieldParserCallback       | Success  |
| 12 | HTTPClient_strerror                 | Success  |

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -143,6 +145,13 @@ def get_args():
             "flags": ["--debug"],
             "action": "store_true",
             "help": "debug output",
+    }, {
+            "flags": ["--tabular-summary-print-fmt"],
+            "type": str,
```
karkhaz, line 150: Not necessary

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -143,6 +145,13 @@ def get_args():
             "flags": ["--debug"],
             "action": "store_true",
             "help": "debug output",
+    }, {
+            "flags": ["--tabular-summary-print-fmt"],
```
karkhaz, line 149: Can I suggest a shorter name? `--tabular-format` or `--summary-format`?

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -200,12 +209,15 @@ def get_proof_dirs(proof_root, proof_list, marker_file):
...
 
-def run_build(litani, jobs, fail_on_proof_failure):
+def run_build(litani, jobs, fail_on_proof_failure, tabular_summary_print_fmt):
     cmd = [str(litani), "run-build"]
     if jobs:
         cmd.extend(["-j", str(jobs)])
     if fail_on_proof_failure:
         cmd.append("--fail-on-pipeline-failure")
+    if tabular_summary_print_fmt:
+        out_file_path = pathlib.Path("run.json")
```
karkhaz, line 219: Why dump `run.json` in the `proofs` directory? It should at least be `.gitignore`d... but can I suggest a different location, like use `tempfile` to create a temporary directory to write into?

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -216,6 +228,12 @@ def run_build(litani, jobs, fail_on_proof_failure):
         logging.critical("Failed to run litani run-build")
         sys.exit(1)
 
+    if tabular_summary_print_fmt:
+        with open(out_file_path) as fp:
+            run_dict = json.load(fp)
```
karkhaz, line 233: Can I suggest more robust error handling here? If for any reason the summary could not be printed out, I think that shouldn't be a fatal error, just print out a warning/error and exit.

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -233,19 +251,29 @@ def get_litani_path(proof_root):
     return proc.stdout.strip()
 
 
-def get_litani_capabilities(litani_path):
-    cmd = [litani_path, "print-capabilities"]
+def get_litani_version(litani_path):
+    cmd = [litani_path, "-V"]
```
karkhaz, line 255: Not sure why this change is necessary. We'll probably revert it in the future as more capabilities are added.

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -269,18 +297,6 @@ def check_uid_uniqueness(proof_dir, proof_uids):
     sys.exit(1)
 
 
-def should_enable_memory_profiling(litani_caps, args):
```
karkhaz, line 272: Why remove this?

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -314,15 +330,45 @@ async def configure_proof_dirs(
...
+        proofs.append([proof, status_pretty_name])
+    statuses = [["Status", "Count"]]
+    for status, count in count_statuses.items():
+        statuses.append([status, count])
+    return [statuses, proofs]
+
+
+def print_proof_results(run_dict, summary_format):
+    statuses, proofs = get_status_and_proof_summaries(run_dict)
+    statuses_summary = tabulate.tabulate(
```
karkhaz, line 354: see above...

```
for table in get_status_and_proof_summaries(run_dict):
    rendered = tabulate.tabulate(
        table["data"], headers="firstrow", tablefmt=summary_format,
        show_index=table["show_index"])
```

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -314,15 +330,45 @@ async def configure_proof_dirs(
...
+        try:
+            count_statuses[status_pretty_name] += 1
+        except KeyError:
+            count_statuses[status_pretty_name] = 1
+        proof = proof_pipeline["name"]
+        proofs.append([proof, status_pretty_name])
+    statuses = [["Status", "Count"]]
+    for status, count in count_statuses.items():
+        statuses.append([status, count])
+    return [statuses, proofs]
```
karkhaz, line 349: Can I suggest:
```
return [{
  "data": statuses,
  "show_index": False,
}, {
  "data": proofs,
  "show_index": True,
}]
```
see below

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -333,17 +379,16 @@ async def main(): # pylint: disable=too-many-locals
...
-            out_symlink = out_prefix / "latest"
-            out_index = out_symlink / "html" / "index.html"
-            cmd.extend([
-                "--output-prefix", str(out_prefix),
-                "--output-symlink", str(out_symlink),
-            ])
-            print(
-                "\nFor your convenience, the output of this run will be symbolically linked to ",
-                out_index, "\n")
+        out_prefix = proof_root / "output"
```
karkhaz, line 382: Again I don't see the need for this change...

test/repo/Makefile
```diff
@@ -81,3 +82,15 @@ compare:
 	@ for html in $$(cd $(REPORT_ROOT2) && find . -name html); do \
 	  diff -rq $(REPORT_ROOT1)/$$html $(REPORT_ROOT2)/$$html; \
 	done
+
+summarize:
+	$(RM) -r $(REPORT_ROOT)
+	cp $(PACKAGE_ROOT)/src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py $(REPO_ROOT)/$(REPO_PROOF_ROOT)
+	cd $(REPO_ROOT)/$(REPO_PROOF_ROOT) && \
```
karkhaz, line 89: I suggest to `cd` on each line, don't do `&& \`. Using the line continuation means that this entire command will be executed in a subshell, and it's difficult to know which command failed if any.

test/repo/Makefile
```diff
@@ -81,3 +82,15 @@ compare:
 	@ for html in $$(cd $(REPORT_ROOT2) && find . -name html); do \
 	  diff -rq $(REPORT_ROOT1)/$$html $(REPORT_ROOT2)/$$html; \
 	done
+
+summarize:
```
karkhaz, line 86: What is this actually testing? I'm not sure that I understand. This just seems to be running the script with the flag?

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -143,6 +145,13 @@ def get_args():
             "flags": ["--debug"],
             "action": "store_true",
             "help": "debug output",
+    }, {
+            "flags": ["--tabular-summary-print-fmt"],
+            "type": str,
+            "choices": tabulate.tabulate_formats,
+            "metavar": "FORMAT",
```
karkhaz, line 152: This gets printed out twice so should be as short as possible, e.g. `F` or `FMT`

karkhaz (changes requested):
Thanks for this. I think this feature should be on by default for interactive users.

@angelonakos what is the most sensible format for interactive users? Could you post some sample output tables for a few different formats?

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: To be honest, looking at the output you posted above, I'm not sure what the value of `show_index` even is. Especially since the proofs themselves aren't in any particular order.

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: In fact, this seems like a flag that users might actually want to tweak. Suggest having a short option too.

test/repo/Makefile
angelonakos: Yes, it's indeed just running the script using the new flag.

I added this target, because it served as a means for me to illustrate this new output without having to ask a reviewer/user to run anything on their machine.

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
angelonakos: As I mentioned in the description of the PR, the check for the associated Litani capability can be removed considering all consumers of the starter-kit use a version of Litani that is >= v1.16.0 and this particular capability was introduced in Litani v1.10.0

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
angelonakos: Indeed this particular change isn't necessary for these changes. I will take this commit out.

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
angelonakos: I perceived the value of the index being that someone looking at the proof-specific table could:
1) Point/refer to a proof at a specific index when interpreting/sharing results
2) Understand the total number of proofs

However, on second thought, I will be removing the index. The total is already available elsewhere.

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
angelonakos: As mentioned above, I will be removing this particular change from the PR.

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
angelonakos: Resolving since reasoning here is the same as with other changes related to no longer checking for Litani capabilities. I'm removing this change.

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
angelonakos: Another reason for removing the index is that rendering the table later breaks in certain platforms.

test/repo/Makefile
angelonakos: I will be removing this target since the Makefile already had a specific intent. I should instead create a separate test.

test/repo/Makefile
angelonakos: Done. I just pushed a separate commit that does that.

test/summarize_test/summarize_test.py
```diff
@@ -0,0 +1,43 @@
...
+        self.run_file = pathlib.Path().cwd() / "sample_run.json"
+
+
+    @patch('builtins.print')
+    def test_print_proof_results(self, mock):
+        summarize.print_proof_results(self.run_file)
+        expected_calls = []
+        expected_calls.append(
+            call(
+                "| Status  | Count |\n"\
```
karkhaz, line 27: No need for backslashes

test/summarize_test/summarize_test.py
```diff
@@ -0,0 +1,43 @@
...
+    def setUp(self):
+        self.run_file = pathlib.Path().cwd() / "sample_run.json"
+
+
+    @patch('builtins.print')
+    def test_print_proof_results(self, mock):
+        summarize.print_proof_results(self.run_file)
+        expected_calls = []
+        expected_calls.append(
+            call(
```
karkhaz, line 26: I'm not going to block on this, but I find un-namespaced function calls to be slightly distracting. I don't see a good reason that this shouldn't be `unittest.mock.call`

test/summarize_test/sample_run.json
```diff
@@ -0,0 +1,187 @@
...
+            "name": "report"
+          }
+        ],
+        "url": "pipelines/pipe-will-succeed",
+        "status": "success",
+        "dependencies_url": "dependencies.svg"
+      }
+    ],
+    "__duration_str": "02m 54s"
+  }
```
karkhaz, line 187: nl

src/cbmc_starter_kit/template-for-repository/proofs/lib/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
```
karkhaz, line 2: I don't think we should have copyright headers on empty files, this doesn't make sense. Amazon only has copy rights on creative output.

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
```diff
@@ -0,0 +1,113 @@
...
+    return [statuses, proofs]
+
+
+def print_proof_results(out_file):
+    """
+    Print 2 strings that summarize the proof results.
+    When printing, each string will render as a GitHub flavored Markdown table.
+    """
+    with open(out_file) as fp:
+        run_dict = json.load(fp)
```
karkhaz: I think this should be inside the `try`, specifically to catch `JSONDecodeError`

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
```diff
@@ -0,0 +1,113 @@
...
+        try:
+            for summary in _get_status_and_proof_summaries(run_dict):
+                print(_get_rendered_table(summary))
+        except Exception as ex:
+            logging.critical("Could not print results. Exception: %s", str(ex))
+            sys.exit(1)
```
karkhaz: Not going to block on this, but I don't think it should be a fatal error

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
```diff
@@ -0,0 +1,113 @@
...
+import logging
+import sys
+
+
+def _get_max_length(column):
+    return max(len(word) for word in column)
+
+
+def _get_list_of_max_length_per_column(data):
+    list_of_max_length_per_column = []
```
karkhaz, line 14: Not going to block on this but I found this a bit difficult to read. What do you think, is this cleaner? (also removes the need for the function above)
```
ret = [len(item) for item in data[0]]
for row in data[1:]:
  for idx, item in enumerate(row):
      ret[idx] = max(ret[idx], len(item))
return ret
```

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
```diff
@@ -0,0 +1,113 @@
...
+    num_columns = len(data[0])
+    for column_idx in range(num_columns):
+        column = [data[row_idx][column_idx] for row_idx in range(num_rows)]
+        max_length = _get_max_length(column)
+        max_length_with_extra_space = max_length + 1
+        list_of_max_length_per_column.append(max_length_with_extra_space)
+    return list_of_max_length_per_column
+
+
+def _get_header(list_of_max_length_per_column, header_names):
```
karkhaz, line 25: I don't understand why this is a separate function. Why not make the header names the first item in the table? They don't appear to be formatted any differently than any of the other items. I would just use `_get_entries` to iterate over all rows including the headers, and then insert a line of dashes in `data[1]`.

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
```diff
@@ -0,0 +1,113 @@
...
+    for idx, word in enumerate(header_names):
+        max_length_of_word_in_col = list_of_max_length_per_column[idx]
+        space_formatted_word = (max_length_of_word_in_col - len(word)) * " "
+        header += "| " + word + space_formatted_word
+    header += "|\n"
+    return header
+
+
+def _get_line_separator_after_header(list_of_max_length_per_column):
+    line_sep = ""
```
karkhaz, line 36: If you do as I suggested above with treating the headers the same as the data, it would make it easier to remove this function. Ideally if you 'render' the first line of the table as a string called `header`, then this would just be `print("-" * len(header)).

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -216,6 +226,10 @@ def run_build(litani, jobs, fail_on_proof_failure):
         logging.critical("Failed to run litani run-build")
         sys.exit(1)
 
+    if summarize:
```
karkhaz, line 229: I cry a little bit whenever I see something that Go's `defer` keyword effortlessly solves...

karkhaz (changes requested):
I left a few comments. But at a higher level... this PR originally represented the table as a list of lists because that's what `tabulate` expects. But since we're not using that any more, I think that the implementation might be a lot cleaner if we don't do that. The trouble with a list of lists is that to render the table, you need a bunch of auxiliary data like the maximum width of the column, which you're computing separately and then pass around as context.

I usually thumb my nose at OOP but I think this code would be cleaner if Table was an object. You'd construct it by giving the headers, and there would be a method to add a row. That method would add the row data to the table but also update the maximum column width, which the table would also store internally. That way, when you've finished adding rows, you can just call `render` on the table and that would be a pretty simple function that iterates over the rows, using `self.width[idx]` which has already been precomputed. I'm not going to block on this as it's a big suggestion and I'm not totally sure that it would make a lot of difference, but I'd at least consider doing this.

test/summarize_test/summarize_test.py
karkhaz: In fact, I marginally prefer
```
call("\n".join([
  # no "\n" in any of the strings
])
```

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
karkhaz: If you want to go overboard you could even replace the inner loop with a comprehension which would save another line, depending on whether you think that would be easier to read.

karkhaz: To be honest I prefer this style, but I realise nested comprehensions are sometimes difficult to read
```
ret = [len(item) for item in data[0]]
return [[max(ret[idx], len(item)) for idx, item in row] for row in data[1:]]
```

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
angelonakos: > ...then this would just be `print("-" * len(header))`.

`print("-" * len(header))` will not work because the resulting table could be something like the following:

```
| Status  | Count |
-------------------
| Fail    | 1    |
| Success | 1    |
```

which could be rendered, if so chosen, as such:


| Status  | Count |
-------------------
| Fail    | 1    |
| Success | 1    |

angelonakos: > ...then this would just be print("-" * len(header))`.

In the case of generating a table that follows the GitHub flavored Markdown spec and is left-aligned appropriately, the line that follows the header and separates the header from the "row entries" needs to be something like:

```
| --------- | ---- | ---------- |
```

where the number of dashes between the pipes is equal to the number of characters inside either the longest "row entry" or the given "header name" for that column.

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
angelonakos: Good point. Removed.

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
angelonakos: I went with your first suggestion which indeed increased readability of the original function.

test/summarize_test/summarize_test.py
angelonakos: Removed the superfluous backslashes. The `"\n".join([ ... ])` bit won't work because each "table string" contains 2 `\n` characters in the very end.

test/summarize_test/summarize_test.py
angelonakos: I used `unittest.mock.call` to be more clear.

karkhaz (approved):

markrtuttle (approved):

PR #122, opened 2022-05-26
Add test for CBMC starter kit and update subprocess command that runs individual proofs

angelonakos:
*Issue #, if available:*

*Description of changes:*

As I was testing my changes that are part of #121 , I noticed that the single cbmc-starter-kit test (i.e. the "test Makefile") had been broken.

This PR:

- introduces a GitHub Actions test workflow
- updates the `make _report` sub-process command to not include `VERBOSE=1`. This was causing the "test Makefie" to fail.

### DIFF
--- a/.github/workflows/makefile-test.yaml
+++ b/.github/workflows/makefile-test.yaml
@@ -1,0 +1,29 @@
+name: Test CBMC starter-kit by using coreHTTP
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+  push:
+    branches: [ master ]
+
+jobs:
+  run-tests:
+    if: "!contains(github.event.pull_request.labels.*.name, 'no-test')"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ '3.8', '3.9', '3.10' ]
+    name: Python ${{ matrix.python-version }}
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/makefile-test.yaml
```diff
@@ -0,0 +1,31 @@
...
+    name: Python ${{ matrix.python-version }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          brew install scdoc mandoc coreutils ninja
```
ronakfof, line 24: Since you are running it on `ubuntu-latest`, I would suggest installing dependencies with `apt` and not `brew` even though it works either way in GitHub Actions.

markrtuttle:
I cannot see from this report how the debug flag broke anything.

.github/workflows/makefile-test.yaml
angelonakos: I decided to go with `brew` because almost all dependencies can be installed within one command. As you suggested `apt` could indeed work as well.

angelonakos:
> I cannot see from this report how the debug flag broke anything.

I force-pushed some changes. Instead of reverting a previous commit, we are now removing the bit that was causing the erroneous behavior.

markrtuttle (approved):

PR #125, opened 2022-06-27
Print summary table even when some proofs fail

karkhaz:
Prior to this commit, passing --fail-on-pipeline-failure would terminate
run-cbmc-proofs.py without printing the summary table if some of the
proofs failed. This commit fixes that.

### DIFF
--- a/src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
+++ b/src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
@@ -219,10 +219,8 @@ def run_build(litani, jobs, fail_on_proof_failure, summarize):
 
     logging.debug(" ".join(cmd))
     proc = subprocess.run(cmd, check=False)
-    if proc.returncode:
-        if fail_on_proof_failure:
-            logging.error("One or more proofs failed")
-            sys.exit(10)
+
+    if proc.returncode and not fail_on_proof_failure:
         logging.critical("Failed to run litani run-build")
         sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
```diff
@@ -219,17 +219,18 @@ def run_build(litani, jobs, fail_on_proof_failure, summarize):
 
     logging.debug(" ".join(cmd))
     proc = subprocess.run(cmd, check=False)
-    if proc.returncode:
-        if fail_on_proof_failure:
-            logging.error("One or more proofs failed")
-            sys.exit(10)
+
+    if proc.returncode and not fail_on_proof_failure:
```
markrtuttle, line 223: Are you sure this is doing what you intend?  If litani fails because goto-cc fails, what should happen?  If litani fails because cbmc failed (eg, internal assertion failure), what should happen?  If litani fails because cbmc ran but has a memory safety finding, what should happen?  Shouldn't the flag being used to address the customer problem affect only the final case?

markrtuttle (changes requested):

src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py
karkhaz: I would think that in all of the cases above, you would want to see a summary table that indicates that one of the proofs failed.

The summary table isn't giving detailed diagnostics about what went wrong. If you want to implement that in the starter kit we can, but that's a whole separate feature and I think time would be better spent improvign the lynx output as you suggest.

What this patch is doing is: if litani crashed, then don't attempt to print a summary because we likely can't read the `run.json` file. In any other circumstance--whether proofs passed or failed and for whatever reason---print out a summary.

markrtuttle (approved):

PR #136, opened 2022-09-22
Remove support for cbmc-viewer 1.0

markrtuttle:
This pull request removes support of deprecated tools (cbmc-batch, cbmc-viewer version 1, and make-source) and adds support for generating reports without coverage: Running "make report" generates a report with property and coverage checking in "report" subdirectory, and running "make report-no-coverage" generates a report with only property checking in "report-no-coverage".

### DIFF
--- a/src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
+++ b/src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
@@ -943,38 +943,6 @@ veryclean: clean
 
 ################################################################
 
-# Rule for generating cbmc-batch.yaml, used by the CI at
-# https://github.com/awslabs/aws-batch-cbmc/
-
-JOB_OS ?= ubuntu16
-JOB_MEMORY ?= 32000
-
-# Proofs that are expected to fail should set EXPECTED to
-# "FAILED" in their Makefile. Values other than SUCCESSFUL
-# or FAILED will cause a CI error.
-EXPECTED ?= SUCCESSFUL
-
-define yaml_encode_options
-	"$(shell echo $(1) | sed 's/ ,/ /g' | sed 's/ /;/g')"
-endef
-
-CI_FLAGS = $(CBMCFLAGS) $(CBMC_FLAG_UNWINDING_ASSERTIONS) $(CHECKFLAGS) $(COVERFLAGS)
-
-cbmc-batch.yaml:
-	@$(RM) $@
-	@echo 'build_memory: $(JOB_MEMORY)' > $@
-	@echo 'cbmcflags: $(strip $(call yaml_encode_options,$(CI_FLAGS)))' >> $@
-	@echo 'coverage_memory: $(JOB_MEMORY)' >> $@
-	@echo 'expected: $(EXPECTED)' >> $@
-	@echo 'goto: $(HARNESS_GOTO).goto' >> $@
-	@echo 'jobos: $(JOB_OS)' >> $@
-	@echo 'property_memory: $(JOB_MEMORY)' >> $@
-	@echo 'report_memory: $(JOB_MEMORY)' >> $@
-
-.PHONY: cbmc-batch.yaml
-
-################################################################
-
 # Run "make echo-proof-uid" to print the proof ID of a proof. This can be
 # used by scripts to ensure that every proof has an ID, that there are
 # no duplicates, etc.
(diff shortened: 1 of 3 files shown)


src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -891,24 +851,24 @@ coverage:
...
+report:
 	@ echo Running 'litani init'
 	$(LITANI) init --project $(PROJECT_NAME)
 	@ echo Running 'litani add-job'
 	$(MAKE) -B _report
 	@ echo Running 'litani build'
 	$(LITANI) run-build
 
+_report_no_coverage: $(PROOFDIR)/report-no-coverage
+report-no-coverage:
```
angelonakos, line 864: Could add:

`* make report-no-coverage: builds the final report without calculating coverage` after L17

inside of `src/cbmc_starter_kit/template-for-repository/proofs/README.md`.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -918,23 +878,20 @@ clean:
...
   _property \
   _report \
-  _report2 \
-  _result \
   clean \
   coverage \
   goto \
   litani-path \
   property \
   report \
```
angelonakos, line 894: You could add `report-no-coverage`

feliperodri (approved):
LGTM.

angelonakos:
You could update `src/cbmc_starter_kit/template-for-repository/proofs/run-cbmc-proofs.py` to enable a user of that script to specify that the reports generated by CBMC viewer for their reports should not contain information concerning coverage.

A `--no-coverage` flag can be added to `get_args()` which will allow you to create the following variable after L390:

```python
report_target = "_report_no_coverage" if args.no_coverage else "_report"
```

and to pass it to the `configure_proof_dirs` function.

Instead of specifying `_report` on L314, you could use `report_target`.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -788,60 +770,38 @@ $(LOGDIR)/coverage.xml: $(HARNESS_GOTO).goto
...
-export VIEWER2_CMD
-
-# Omit logs/source.json from report generation until make-sources
-# works correctly with Makefiles that invoke the compiler with
-# mutliple source files at once.
-$(PROOFDIR)/report: $(LOGDIR)/result.xml $(LOGDIR)/property.xml $(LOGDIR)/coverage.xml
+$(PROOFDIR)/report-no-coverage: $(LOGDIR)/result.xml $(LOGDIR)/property.xml
 	$(LITANI) add-job \
-	  --command "$$VIEWER2_CMD" \
+	  --command " $(VIEWER) \
```
karkhaz, line 792: This seems like a lot of duplicated code just to change one flag. I think there should just be one recipe for report, and the coverage flag should be defined out if the user asks for it. Something like this (not tested):
```
# early in the file
VIEWER_COVER_FLAG ?= "--coverage $(LOGDIR)/coverage.xml"

report-no-coverage:
  $(MAKE) VIEWER_COVER_FLAG="" _report

# rule for report
# ...
        --command " $(VIEWER) \
	    --result $(LOGDIR)/result.xml \
	    $(VIEWER_COV_FLAG) \
	    --property $(LOGDIR)/property.xml \
```

karkhaz: (and have a different variable for the output directory if it also needs to be different, etc)

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
markrtuttle: Done.  Actually rewrite the text to be more readabale.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
markrtuttle: Added both no coverage targets.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
markrtuttle: I've made these changes.

markrtuttle:
This has been tested on the HTTP repository, and I've demonstrated that the sets of commands issued by the master version and this version are the same for the report target.

angelonakos (approved):

PR #149, opened 2022-11-22
Remove flags from update submcommand

angelonakos:
*Issue #, if available:*

*Description of changes:*

The 2 deleted flags would remove the CBMC starter kit, Litani submodules, if a project had configured their C project to use those 2 as submodules.

As a result of this change, all users of the CBMC starter kit will have to install it (and Litani) as a package.


karkhaz (approved):
Awesome, thanks for this

PR #151, opened 2022-11-28
Fix test that hasn't been running in CI for recent PRs

angelonakos:
*Issue #, if available:*

*Description of changes:*

The relevant check has been failing in CI for all open PRs.

Edit: Due to the fact that this test was hanging in CI, certain proofs have been removed in an effort to avert this behavior.

### DIFF
--- a/.github/workflows/makefile-test.yaml
+++ b/.github/workflows/makefile-test.yaml
@@ -19,29 +19,31 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
+      - name: Install latest CBMC
+        id: install_latest_cbmc
+        shell: bash
         run: |
-          get_latest_release() {
-            curl --silent "https://api.github.com/repos/$1/releases/latest" |
-            jq -r '.tag_name'
-          }
-          cbmc_latest_release=$(get_latest_release diffblue/cbmc)
-          litani_latest_release=$(get_latest_release awslabs/aws-build-accumulator)
-
-          curl -o cbmc.deb -L \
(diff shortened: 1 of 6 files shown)


.github/workflows/makefile-test.yaml
```diff
@@ -19,29 +19,31 @@ jobs:
...
-                                            ./cbmc.deb \
-                                            ./litani.deb \
-                                            scdoc \
-                                            mandoc
-
-          python3 -m pip install pyyaml jinja2 cbmc-viewer
-
+          # Search within 5 most recent releases for latest available package
+          CBMC_REL="https://api.github.com/repos/diffblue/cbmc/releases?page=1&per_page=5"
+          CBMC_DEB=$(curl -s $CBMC_REL | jq -r '.[].assets[].browser_download_url' | grep -e 'ubuntu-20.04' | head -n 1)
```
karkhaz, line 28: Suggest `curl -o` to save to a file so that the script fails early if this doesn't work. Then define the variable separately. I also recommend to check that the variable is non-empty, because you're `curl`ing it 2 lines down so it's good to fail early if the URL doesn't exist for any reason.

karkhaz (changes requested):

.github/workflows/makefile-test.yaml
angelonakos: If a failure occurs when trying to determine `CBMC_DEB`, then that variable will be empty.
As a result, the workflow will fail when `CBMC_ARTIFACT_NAME` is created. A failure here means that fetching an Ubuntu 20.04 download URI within the latest 5 releases didn't work. Therefore, an "early exit" does take place as the second `curl` 2 lines down, as you mentioned, won't even be executed.

karkhaz (approved):

PR #156, opened 2022-12-12
Print versions of all tools to front Litani page

karkhaz:
The Litani dashboard for every proof run now includes a table indicating the versions of CBMC, CBMC Viewer, and other proof tools that were used in this run.

### DIFF
--- a/src/cbmc_starter_kit/print_tool_versions.py
+++ b/src/cbmc_starter_kit/print_tool_versions.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python3
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+
+
+import logging
+import pathlib
+import shutil
+import subprocess
+
+
+_TOOLS = [
+    "cadical",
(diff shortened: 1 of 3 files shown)


karkhaz:
Here's what it looks like:

Screenshot 2022-12-12 at 22 42 23

src/cbmc_starter_kit/print_tool_versions.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+import logging
+import pathlib
+import shutil
+import subprocess
+
+
+_TOOLS = [
+    "cadical",
```
feliperodri, line 14: Could we also print minisat version since we are printing cadical?

feliperodri (approved):

src/cbmc_starter_kit/print_tool_versions.py
karkhaz: I don't think minisat is installed as an executable, CBMC statically links it.

I need some kind of executable command that prints out the version number. I don't think we have anything like this for minisat. One path forward might be to add a flag to CBMC that prints out the minisat version, if this is important

PR #157, opened 2022-12-13
Update loop unwinding order

feliperodri:
# Change overview
We update the Makefile.common so that :
- When function contracts with DFCC is used, the CPROVER library must be linked and all loops that need unwinding be unwound  before contract instrumentation
- When the old function contracts or loop contracts are used the library must be linked after contract instrumentation. Loops in user code must be unwound before contract instrumentation and loops in CPROVER library functions unwound after contract instrumentation
- When no contracts are used, the CPROVER library and loop unwinding are performed a the last step of the build when cbmc is invoked.

# CPROVER Library flags management

The flags configuring the CPROVER library (i.e., CBMC_OPT_CONFIG_LIBRARY) must be passed at any step where the library is be loaded.

- When linking with goto instrument `goto-instrument --add-library $(CBMC_OPT_CONFIG_LIBRARY) ...`
- When applying contracts with DFCC `goto-instrument  $(CBMC_OPT_CONFIG_LIBRARY) --dfcc harness...` because the behavior  of the built-in `__CPROVER_is_fresh` is also configured by malloc failure modes
- When linking to the library at analysis time `cbmc $(CBMC_OPT_CONFIG_LIBRARY) ...`

# Loop unwinding

Loop unwinding is parameterized by the following  envs:
- `UNWINDSET` : loops to unwind in user code (i.e., proof and project code);
- `CPROVER_LIBRARY_UNWINDSET` : loops to unwind in CPROVER library functions; 
- `DEFAULT_UNWIND`: `--unwind 1` (sets a default unwinding depth for all loops of a model)
- `STRING_ABSTRACTION`: --string-abstraction (abstracts away loops in `string.h` functions when set)

Loops are handled as follows:
- loops in user code: 
  - The subset of user loops to manually unwind is specified in `UNWINDSET`;
  - Loops that have loop contracts and are not listed in `UNWINDSET` will be abstracted (i.e., `--apply-loop-contracts`);
  - The rest of the loops is handled using `DEFAULT_UNWIND`
- loops in CPROVER library.
  - The subset of CPROVER library loops to unwind is specified in `CPROVER_LIBRARY_UNWINDSET`;
  - The rest of the loops is handled using `DEFAULT_UNWIND` or by using `STRING_ABSTRACTION` (e.g., `strlen`);
- loops in contract instrumentation functions on DFCC mode. These loops *must* be dynamically unwound during analysis;

## When no contracts are used

1. Early unwind. We call `goto-instrument $(UNWINDSET) $(CBMC_FLAG_UNWINDING_ASSERTIONS)` to unwind specific subsets of loops in user code.

2. Linking CPROVER library and loop unwinding for all remaining loops happens at the same time when invoking CBMC (i.e., property checking, coverage checking and printing safety properties).

```Makefile
cbmc $(CBMC_VERBOSITY) $(CBMCFLAGS) $(CHECKFLAGS)
      $(CBMC_OPT_CONFIG_LIBRARY)
      $(DEFAULT_UNWIND) $(CPROVER_LIBRARY_UNWINDSET) $(CBMC_FLAG_UNWINDING_ASSERTIONS)
      --trace --xml-ui
```

## When old function contracts or loop contracts are used

After restricting function pointers:
1. Early unwind. We call `goto-instrument $(UNWINDSET) $(CBMC_FLAG_UNWINDING_ASSERTIONS)` to unwind specific subsets of loops in user code.
3. Contracts instrumentation. We call `goto-instrument --apply-loop-contracts --enforce-contract ...` to instrument the model. This abstracts loops that have contracts, as well as loops that are in functions that get replaced by contracts.
4. Clean up global variables. We call `goto-instrument --slice-global-init ...`
5. Clean up unused functions. We call `goto-instrument --drop-unused-functions ...`
8. Invoke CBMC.

```Makefile
cbmc $(CBMC_VERBOSITY) $(CBMCFLAGS) $(CHECKFLAGS)
      $(CBMC_OPT_CONFIG_LIBRARY)
      $(DEFAULT_UNWIND) $(CPROVER_LIBRARY_UNWINDSET) $(CBMC_FLAG_UNWINDING_ASSERTIONS)
      --trace --xml-ui
```

## When function contracts with DFCC are used

After restricting function pointers:
1. Link CPROVER library. We call `goto-instrument --add-library  $(CBMC_OPT_CONFIG_LIBRARY) ...`. to link the model to the CPROVER standard library;
6. Early unwind. We call `goto-instrument $(CPROVER_LIBRARY_UNWINDSET) $(UNWINDSET) $(DEFAULT_UNWIND) $(CBMC_FLAG_UNWINDING_ASSERTIONS)`;
7. Contracts instrumentation with DFCC. We call `goto-instrument $(CBMC_OPT_CONFIG_LIBRARY) --dfcc harness ...`. The loops found in `__CPROVER_contracts`  *must* be dynamically unwound during analysis which rules out using `--unwind 1` with CBMC. DFCC mode also havocs the global state and drops unused functions;
8. Invoke CBMC.

```Makefile
cbmc $(CBMC_VERBOSITY) $(CBMCFLAGS) $(CHECKFLAGS)
      $(CBMC_OPT_CONFIG_LIBRARY)
      $(CBMC_FLAG_UNWINDING_ASSERTIONS)
      --trace --xml-ui
```

Signed-off-by: Felipe R. Monteiro <felisous@amazon.com>

### DIFF
--- a/src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
+++ b/src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
@@ -252,12 +252,12 @@ CBMC_FLAG_SIGNED_OVERFLOW_CHECK ?= --signed-overflow-check
 CBMC_FLAG_UNDEFINED_SHIFT_CHECK ?= --undefined-shift-check
 CBMC_FLAG_UNSIGNED_OVERFLOW_CHECK ?= --unsigned-overflow-check
 CBMC_FLAG_UNWINDING_ASSERTIONS ?= --unwinding-assertions
-CBMC_FLAG_UNWIND ?= --unwind 1
+CBMC_DEFAULT_UNWIND ?= --unwind 1
 CBMC_FLAG_FLUSH ?= --flush
 
 # CBMC flags used for property checking and coverage checking
 
-CBMCFLAGS += $(CBMC_UNWINDSET) $(CBMC_FLAG_FLUSH)
+CBMCFLAGS += $(CBMC_FLAG_FLUSH)
 
 # CBMC flags used for property checking
 
@@ -303,6 +303,9 @@ COMPILE_FLAGS ?= -Wall
 LINK_FLAGS ?= -Wall
 EXPORT_FILE_LOCAL_SYMBOLS ?= --export-file-local-symbols
 
+# During instrumentation, it adds models of C library functions
+ADD_LIBRARY_FLAG := --add-library
+
 # Preprocessor include paths -I...
 INCLUDES ?=
 
@@ -341,9 +344,9 @@ UNWINDSET ?=
 # contracts).  To satisfy this requirement, it may be necessary to
 # unwind some loops before the function contract and loop invariant
 # transformations are applied to the goto program.  This variable
-# EARLY_UNWINDSET is identical to UNWINDSET, and we assume that the
-# loops mentioned in EARLY_UNWINDSET and UNWINDSET are disjoint.
-EARLY_UNWINDSET ?=
+# CPROVER_LIBRARY_UNWINDSET is identical to UNWINDSET, and we assume that the
+# loops mentioned in CPROVER_LIBRARY_UNWINDSET and UNWINDSET are disjoint.
+CPROVER_LIBRARY_UNWINDSET ?=
 
 # CBMC function removal (Normally set set in the proof Makefile)
 #
@@ -393,10 +396,15 @@ CBMC_TIMEOUT ?= 21600
 # Replace all uses of char * by a struct that carries that string,
 # and also the underlying allocation and the C string length.
 STRING_ABSTRACTION ?=
+ifdef STRING_ABSTRACTION
+  ifneq ($(strip $(STRING_ABSTRACTION)),)
+    CBMC_STRING_ABSTRACTION := --string-abstraction
+  endif
+endif
 
 # Optional configuration library flags
 OPT_CONFIG_LIBRARY ?=
-CBMC_OPT_CONFIG_LIBRARY := $(CBMC_FLAG_MALLOC_MAY_FAIL) $(CBMC_FLAG_MALLOC_FAIL_NULL) $(NONDET_STATIC) $(STRING_ABSTRACTION)
+CBMC_OPT_CONFIG_LIBRARY := $(CBMC_FLAG_MALLOC_MAY_FAIL) $(CBMC_FLAG_MALLOC_FAIL_NULL) $(CBMC_STRING_ABSTRACTION)
 
 # Proof writers could add function contracts in their source code.
 # These contracts are ignored by default, but may be enabled in two distinct
@@ -418,13 +426,15 @@ CBMC_CHECK_FUNCTION_CONTRACTS_REC := $(patsubst %,--enforce-contract-rec %, $(CH
 USE_FUNCTION_CONTRACTS ?=
 CBMC_USE_FUNCTION_CONTRACTS := $(patsubst %,--replace-call-with-contract %, $(USE_FUNCTION_CONTRACTS))
 
+CODE_CONTRACTS := $(CHECK_FUNCTION_CONTRACTS)$(USE_FUNCTION_CONTRACTS)$(APPLY_LOOP_CONTRACTS)
+
 # Proof writers may also apply function contracts using the Dynamic Frame
 # Condition Checking (DFCC) mode. For more information on DFCC,
 # please see https://diffblue.github.io/cbmc/contracts-dev-spec-dfcc.html.
-DFCC_MODE ?=
-ifdef DFCC_MODE
-  ifneq ($(strip $(DFCC_MODE)),)
-    CBMC_DFCC_MODE := --dfcc $(HARNESS_ENTRY) $(CBMC_OPT_CONFIG_LIBRARY)
+USE_DYNAMIC_FRAMES ?=
+ifdef USE_DYNAMIC_FRAMES
+  ifneq ($(strip $(USE_DYNAMIC_FRAMES)),)
+    CBMC_USE_DYNAMIC_FRAMES := $(CBMC_OPT_CONFIG_LIBRARY) --dfcc $(HARNESS_ENTRY) $(CBMC_CHECK_FUNCTION_CONTRACTS_REC)
   endif
 endif
 
@@ -489,14 +499,18 @@ DEFINES += -DCBMC_OBJECT_BITS=$(CBMC_OBJECT_BITS)
 DEFINES += -DCBMC_MAX_OBJECT_SIZE="(SIZE_MAX>>(CBMC_OBJECT_BITS+1))"
 
 # CI currently assumes cbmc invocation has at most one --unwindset
+
+# UNWINDSET is designed for user code (i.e., proof and project code)
 ifdef UNWINDSET
   ifneq ($(strip $(UNWINDSET)),)
     CBMC_UNWINDSET := --unwindset $(subst $(SPACE),$(COMMA),$(strip $(UNWINDSET)))
   endif
 endif
-ifdef EARLY_UNWINDSET
-  ifneq ($(strip $(EARLY_UNWINDSET)),)
-    CBMC_EARLY_UNWINDSET := --unwindset $(subst $(SPACE),$(COMMA),$(strip $(EARLY_UNWINDSET)))
+
+# CPROVER_LIBRARY_UNWINDSET is designed for CPROVER library functions
+ifdef CPROVER_LIBRARY_UNWINDSET
+  ifneq ($(strip $(CPROVER_LIBRARY_UNWINDSET)),)
+    CBMC_CPROVER_LIBRARY_UNWINDSET := --unwindset $(subst $(SPACE),$(COMMA),$(strip $(CPROVER_LIBRARY_UNWINDSET)))
   endif
 endif
 
@@ -638,30 +652,19 @@ $(HARNESS_GOTO)1.goto: $(PROOF_GOTO)1.goto $(PROJECT_GOTO)2.goto
 
 # Restrict function pointers
 $(HARNESS_GOTO)2.goto: $(HARNESS_GOTO)1.goto
-ifneq ($(strip $(RESTRICT_FUNCTION_POINTER)),)
 	$(LITANI) add-job \
 	  --command \
-	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) $(CBMC_RESTRICT_FUNCTION_POINTER) $^ $@' \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) $(CBMC_RESTRICT_FUNCTION_POINTER) --remove-function-pointers $^ $@' \
 	  --inputs $^ \
 	  --outputs $@ \
 	  --stdout-file $(LOGDIR)/restrict_function_pointer-log.txt \
 	  --pipeline-name "$(PROOF_UID)" \
 	  --ci-stage build \
 	  --description "$(PROOF_UID): restricting function pointers in project sources"
-else
-	$(LITANI) add-job \
-	  --command 'cp $^ $@' \
-	  --inputs $^ \
-	  --outputs $@ \
-	  --stdout-file $(LOGDIR)/restrict_function_pointer-log.txt \
-	  --pipeline-name "$(PROOF_UID)" \
-	  --ci-stage build \
-	  --description "$(PROOF_UID): not restricting function pointers in project sources"
-endif
 
 # Fill static variable with unconstrained values
 $(HARNESS_GOTO)3.goto: $(HARNESS_GOTO)2.goto
-ifneq ($(strip $(DFCC_MODE)),)
+ifneq ($(strip $(CODE_CONTRACTS)),)
 	$(LITANI) add-job \
 	  --command 'cp $^ $@' \
 	  --inputs $^ \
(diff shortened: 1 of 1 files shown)


remi-delmas-3000 (approved):

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -682,75 +685,76 @@ else
 	  --description "$(PROOF_UID): setting static variables to nondet"
 endif
 
-# Omit unused functions (sharpens coverage calculations)
+# Omit initialization of unused global variables (reduces problem size)
```
remi-delmas-3000, line 688: we should move this after contract instrumentation, same step as "omit unused functions"

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -682,75 +685,76 @@ else
...
-	  --stdout-file $(LOGDIR)/drop_unused_functions-log.txt \
+	  --stdout-file $(LOGDIR)/slice_global_inits-log.txt \
 	  --pipeline-name "$(PROOF_UID)" \
 	  --ci-stage build \
-	  --description "$(PROOF_UID): dropping unused functions"
+	  --description "$(PROOF_UID): slicing global initializations"
 endif
 
-# Omit initialization of unused global variables (reduces problem size)
+# Early unwind loops (e.g., for C library functions)
```
remi-delmas-3000, line 711: Suggested comment:  "link CPROVER library and unwind all loops (DFCC on) or unwind loops in user code only (DFCC off)"

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -765,6 +769,10 @@ $(HARNESS_GOTO).goto: $(HARNESS_GOTO)7.goto
 ################################################################
 # Targets to run the analysis commands
 
+ifeq ($(strip $(DFCC_MODE)),)
```
remi-delmas-3000, line 772: Yes !

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -682,75 +685,76 @@ else
...
+	  --description "$(PROOF_UID): unwinding C library loops during instrumentation"
 endif
 
-# Early unwind loops (e.g., for functions annotated with contracts)
+# Replace function contracts, check function contracts, instrument for loop contracts
 $(HARNESS_GOTO)6.goto: $(HARNESS_GOTO)5.goto
 	$(LITANI) add-job \
 	  --command \
-	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) $(CBMC_EARLY_UNWINDSET) $(CBMC_FLAG_UNWINDING_ASSERTIONS) $^ $@' \
+	    '$(GOTO_INSTRUMENT) $(CBMC_DFCC_MODE) $(CBMC_VERBOSITY) $(CBMC_CHECK_FUNCTION_CONTRACTS) $(CBMC_CHECK_FUNCTION_CONTRACTS_REC) $(CBMC_USE_FUNCTION_CONTRACTS) $(CBMC_APPLY_LOOP_CONTRACTS) $^ $@' \
```
remi-delmas-3000, line 739: we are missing the `--nondet-static-exclude VAR` flags here, cf.`$(NONDET_STATIC)`

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -682,75 +685,76 @@ else
...
+	  --stdout-file $(LOGDIR)/unwind_loops-log.txt \
 	  --pipeline-name "$(PROOF_UID)" \
 	  --ci-stage build \
-	  --description "$(PROOF_UID): not slicing global initializations"
+	  --description "$(PROOF_UID): unwinding all loops during instrumentation"
 else
 	$(LITANI) add-job \
 	  --command \
-	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) --slice-global-inits $^ $@' \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) $(ADD_LIBRARY_FLAG) $(CBMC_EARLY_UNWINDSET) $(CBMC_FLAG_UNWINDING_ASSERTIONS) $^ $@' \
```
remi-delmas-3000, line 726: When DFCC is not used, we must not add the CPROVER library at this stage, instead it needs to be added after instrumentation has happened, when calling `cbmc`

remi-delmas-3000 (changes requested):
There is an off-by one difference in interpretation of the unwinding bound value between cbmc and goto-instrument. Since the UNWINDSET is now passed to either of them depending on wether DFCC is used, do you think we should automatically adjust the value so it means the same thing in both cases ? 

do we also need to propagate the changes to  the following target ?
```
$(LOGDIR)/coverage.xml: $(HARNESS_GOTO).goto
```

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
feliperodri: Implemented as suggested.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
feliperodri: I've updated the comment.

feliperodri:
> There is an off-by one difference in interpretation of the unwinding bound value between cbmc and goto-instrument. Since the UNWINDSET is now passed to either of them depending on wether DFCC is used, do you think we should automatically adjust the value so it means the same thing in both cases ?
> 
> do we also need to propagate the changes to the following target ?
> 
> ```
> $(LOGDIR)/coverage.xml: $(HARNESS_GOTO).goto
> ```

So I wrote this little program.
```c
int main()
{
  int i = 0;
  int j = 10;
  while(i < j)
  {
    i++;
  }
}
```
If we unroll the loop and verify it with CBMC as follow, we get an assertion violation.
```
cbmc main.c --unwinding-assertions --unwind 10
```
If we unroll the loop using `goto-instrument` using the same bound (as follow), the verification succeeds.
```
goto-cc main.c -o main; goto-instrument main main-inst --unwind 10 --unwinding-assertions; cbmc main-inst
```

**Takeaway** Users now add +1 to their bound, if our change move unrolling to `goto-instrument`, there should be no unwinding failures. I’d document this off by 1 difference, but not adjust the value.

feliperodri:
@remi-delmas-3000 I tested the new `Makefile.common` with the following proofs:
1. Proofs from `coreHTTP` and `coreJSON` without any contracts;
2. Proofs from `coreJSON` with contracts on DFCC mode;
3. Proofs from `coreJSON` off DFCC mode;
4. Proofs from AWS C Common with loop contracts;

remi-delmas-3000 (approved):

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -341,9 +344,9 @@ UNWINDSET ?=
 # contracts).  To satisfy this requirement, it may be necessary to
 # unwind some loops before the function contract and loop invariant
 # transformations are applied to the goto program.  This variable
-# EARLY_UNWINDSET is identical to UNWINDSET, and we assume that the
-# loops mentioned in EARLY_UNWINDSET and UNWINDSET are disjoint.
-EARLY_UNWINDSET ?=
+# CPROVER_LIBRARY_UNWINDSET is identical to UNWINDSET, and we assume that the
+# loops mentioned in CPROVER_LIBRARY_UNWINDSET and UNWINDSET are disjoint.
+CPROVER_LIBRARY_UNWINDSET ?=
```
karkhaz, line 349: 🍓 This changes a variable that users can set in their Makefiles and thus is likely to break somebody's proof. Why is this change needed?

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -393,10 +396,15 @@ CBMC_TIMEOUT ?= 21600
 # Replace all uses of char * by a struct that carries that string,
 # and also the underlying allocation and the C string length.
 STRING_ABSTRACTION ?=
+ifdef STRING_ABSTRACTION
```
karkhaz, line 399: 🥝 Why do we need both `ifdef` and `ifneq`? I think `ifneq` is sufficient and consistent with what we do in the rest of the file

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -252,12 +252,12 @@ CBMC_FLAG_SIGNED_OVERFLOW_CHECK ?= --signed-overflow-check
 CBMC_FLAG_UNDEFINED_SHIFT_CHECK ?= --undefined-shift-check
 CBMC_FLAG_UNSIGNED_OVERFLOW_CHECK ?= --unsigned-overflow-check
 CBMC_FLAG_UNWINDING_ASSERTIONS ?= --unwinding-assertions
-CBMC_FLAG_UNWIND ?= --unwind 1
+CBMC_DEFAULT_UNWIND ?= --unwind 1
```
karkhaz, line 255: 🍓

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -303,6 +303,9 @@ COMPILE_FLAGS ?= -Wall
 LINK_FLAGS ?= -Wall
 EXPORT_FILE_LOCAL_SYMBOLS ?= --export-file-local-symbols
 
+# During instrumentation, it adds models of C library functions
+ADD_LIBRARY_FLAG ?= --add-library
```
karkhaz, line 307: If this is supposed to be switched on all the time, please use `:=`.

If users should be able to switch it off, please do
```
  ifeq ($(strip $(NO_ADD_LIBRARY_FLAG)),)
    CBMC_ADD_LIBRARY_FLAG := --add-library
  else
    CBMC_ADD_LIBRARY_FLAG := 
  endif
```

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -418,13 +426,15 @@ CBMC_CHECK_FUNCTION_CONTRACTS_REC := $(patsubst %,--enforce-contract-rec %, $(CH
...
+
 # Proof writers may also apply function contracts using the Dynamic Frame
 # Condition Checking (DFCC) mode. For more information on DFCC,
 # please see https://diffblue.github.io/cbmc/contracts-dev-spec-dfcc.html.
-DFCC_MODE ?=
-ifdef DFCC_MODE
-  ifneq ($(strip $(DFCC_MODE)),)
-    CBMC_DFCC_MODE := --dfcc $(HARNESS_ENTRY) $(CBMC_OPT_CONFIG_LIBRARY)
+USE_DYNAMIC_FRAMES ?=
+ifdef USE_DYNAMIC_FRAMES
```
karkhaz, line 435: 🥝

karkhaz (changes requested):

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
jimgrundy: Is this orthogonal to the changes needed for contracts and DFCC? If so, perhaps it could be offered as a separate PR with explanation for the motivation?

jimgrundy: I believe Mark had a similar concern, so if it could be factored out it might progress more smoothly.

feliperodri: Nobody uses this variables in their proofs today. This change is necessary cause we may need to unroll loops for CPROVER library (e.g., models for `strlen` and `memcpy`) at different stages depending whether users will use contracts or not. For instance, using dynamic frames, users need to unroll everything during instrumentation; however, if no contracts are used, then they may unroll these loops when invoking CBMC for property checking.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
feliperodri: Same answer as 🍓. Plus, this is an internal variable that it's only used in the `Makefile.common`.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
karkhaz: OK

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
feliperodri: I was following the same pattern used for other variables. I thought the idea was to avoid the comparing and `strip` at all if the variable is not even defined.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
feliperodri: Same answer for 🥝.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
feliperodri: For example,
```Makefile
ifdef RESTRICT_FUNCTION_POINTER
  ifneq ($(strip $(RESTRICT_FUNCTION_POINTER)),)
    CBMC_RESTRICT_FUNCTION_POINTER := $(patsubst %,--restrict-function-pointer %, $(RESTRICT_FUNCTION_POINTER))
  endif
endif
```

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
feliperodri: Implemented using `:=` as suggested.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
karkhaz: It actually was never used in Makefile.common before. So we are now passing `--unwind 1` where we were not before.

I'm very confused.

How did the proofs ever work when --unwind 1 was not being passed to them?

It seems that there is now a risk that we break proofs by newly passing --unwind 1 because the proofs did not have that restriction before, and so adding this unwind bound may cause an unwinding assertion to fail?

karkhaz (changes requested):

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
feliperodri: @karkhaz It was always used before. This recent [commit](https://github.com/model-checking/cbmc-starter-kit/commit/36acd4385cd01a3ee449b334bdf9ac472146a0b1) that removed from our `Makefile.common`.

feliperodri: To be more specific, this [commit](https://github.com/model-checking/cbmc-starter-kit/pull/150/commits/691274ad64b2b4146f1e3116ea0cf832ae7b046b).

karkhaz (approved):

jimgrundy (approved):
I'm content now.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
remi-delmas-3000: In earlier versions of the starter kit (up to the one before this PR)

The `--unwind 1` flag was passed to `cbmc` trough in `CBMC_FLAGS`

```
CBMC_FLAG_UNWIND ?= --unwind 1
...
CBMCFLAGS += $(CBMC_FLAG_UNWIND) $(CBMC_UNWINDSET) $(CBMC_FLAG_FLUSH)
```

And the unwinding assertions flag was passed together with CHECK_FLAGS
```
CBMC_FLAG_UNWINDING_ASSERTIONS ?= --unwinding-assertions
...
CHECKFLAGS += $(CBMC_FLAG_UNWINDING_ASSERTIONS)
```

So all proofs were running using `--unwind 1` as default unless they were already processed using `UNWINDSET`.   

In this PR this is still how things happen when no contracts are used. 

When old contracts are used, user loops are unwound before contract instrumentation, and CPROVER library loops are unwound by cbmc at the end, and -- unwind 1 is also given to cbmc.

When DFCC contracts are used, the user loops and CPROVER library loops are unwound before contract instrumentation and --unwind 1 is applied as well before contract instrumentation, which results in all loops being bounded just like when no contracts are used or when old contracts are used, except for the loops introduced by the dynamic frames instrumentation which are dynamically unwound during symex (and always terminate).

remi-delmas-3000 (approved):

tautschnig (commented):
See my comments for specific concerns. Overall, however, I am really concerned about the complexity we have arrived at. In my opinion, we should have completely separate rule sets for the three modes (no contracts, contracts, dfcc) by just putting the rules in three different Makefiles. Depending on the flags set we would then source only one of the three files.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -682,78 +685,100 @@ else
...
 
-# Omit unused functions (sharpens coverage calculations)
+# Link CPROVER library if DFCC mode is on
 $(HARNESS_GOTO)4.goto: $(HARNESS_GOTO)3.goto
-ifneq ($(strip $(DFCC_MODE)),)
+ifneq ($(strip $(USE_DYNAMIC_FRAMES)),)
 	$(LITANI) add-job \
-	  --command 'cp $^ $@' \
+	  --command \
+	    '$(GOTO_INSTRUMENT) $(CBMC_VERBOSITY) $(ADD_LIBRARY_FLAG) $(CBMC_OPT_CONFIG_LIBRARY) $^ $@' \
```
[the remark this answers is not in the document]
remi-delmas-3000: yes they are still going to work as intended. When using contracts for function pointers, function pointer restrictions are expected to contain the `function_pointer -> contract_function` mapping explicitly. This will have to be changed later when function pointers get resolved during symex.

feliperodri:
> See my comments for specific concerns. Overall, however, I am really concerned about the complexity we have arrived at. In my opinion, we should have completely separate rule sets for the three modes (no contracts, contracts, dfcc) by just putting the rules in three different Makefiles. Depending on the flags set we would then source only one of the three files.

@tautschnig I opened https://github.com/model-checking/cbmc-starter-kit/issues/160 issue, so we can refactor `Makefile.common` on a separate PR to reduce its complexity.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -489,14 +499,18 @@ DEFINES += -DCBMC_OBJECT_BITS=$(CBMC_OBJECT_BITS)
 DEFINES += -DCBMC_MAX_OBJECT_SIZE="(SIZE_MAX>>(CBMC_OBJECT_BITS+1))"
 
 # CI currently assumes cbmc invocation has at most one --unwindset
+
+# UNWINDSET is design for user code (i.e., proof and project code)
```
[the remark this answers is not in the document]
feliperodri: Fixed typo.

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -489,14 +499,18 @@ DEFINES += -DCBMC_OBJECT_BITS=$(CBMC_OBJECT_BITS)
...
 ifdef UNWINDSET
   ifneq ($(strip $(UNWINDSET)),)
     CBMC_UNWINDSET := --unwindset $(subst $(SPACE),$(COMMA),$(strip $(UNWINDSET)))
   endif
 endif
-ifdef EARLY_UNWINDSET
-  ifneq ($(strip $(EARLY_UNWINDSET)),)
-    CBMC_EARLY_UNWINDSET := --unwindset $(subst $(SPACE),$(COMMA),$(strip $(EARLY_UNWINDSET)))
+
+# CPROVER_LIBRARY_UNWINDSET is design for CPROVER library functions
```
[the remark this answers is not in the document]
feliperodri: Fixed typo.

PR #159, opened 2022-12-14
[GitHub Actions CI] Enable uploading and online browsing of CBMC proof artifacts

angelonakos:
Issue #, if available:

Description of changes:

This PR is a followup to https://github.com/model-checking/cbmc-starter-kit/pull/152

The GitHub Actions workflow, which handles the execution of CBMC proofs in CI, can now optionally upload results to S3 and make them available for online browsing via CloudFront (assuming that the developer, who is responsible for setting up the CI, will provide their AWS account ID during setup).

Just like before CBMC proofs are executed inside of the GitHub Actions runner. Artifacts from the execution of CBMC proofs are uploaded to a bucket in S3. Due to fact that this bucket is "associated" to a CloudFront distribution, users are able to browse online all related artifacts. A user only needs to click on the CBMC proof report link, which is present both in the logs of the GitHub Actions workflow and in the "summary" page of the corresponding GitHub Actions job.

[This PR in my fork of aws/s2n-tls](https://github.com/angelonakos/s2n-tls/pull/17) serves as a demonstration of the changes here.

### DIFF
--- a/src/cbmc_starter_kit/setup_ci.py
+++ b/src/cbmc_starter_kit/setup_ci.py
@@ -5,20 +5,32 @@
 
 """Set up the AWS infrastructure to run proofs in CI."""
 
+import json
 import logging
 import shutil
-from argparse import RawDescriptionHelpFormatter
+from argparse import ArgumentTypeError, RawDescriptionHelpFormatter
 
+import boto3
+import botocore.exceptions
+import git
 
 from cbmc_starter_kit import arguments, repository, util
 
 ################################################################
+
+def is_valid_aws_account_id(aws_account_id):
+    if not aws_account_id or not (aws_account_id.isdigit() and len(aws_account_id) == 12):
+        raise ArgumentTypeError(
+            "AWS account ID is a 12-character string consisting only of digits")
+    return aws_account_id
 
 def parse_arguments():
     """Parse arguments for cbmc-starter-kit-setup-ci command"""
     desc = """
     Copy a GitHub Action workflow to `.github/workflows` in this repository,
     which runs CBMC proofs on every push event.
+    If your project is a public GitHub repo, you can elect to provision
+    infrastructure in AWS, such that you can browse CI artifacts online.
     """
     eplg = """
     The most recently released and available version of a tool (like CBMC, CBMC
@@ -41,6 +53,14 @@ def parse_arguments():
                 The GitHub-hosted runner (operating on Ubuntu 20.04) that will
                 run CBMC proofs in GitHub Actions. If your repo has a large
                 runner available, you can specify it here. default: %(default)s"""
+        }, {
+        'flag': '--aws-account-id',
+        'type': is_valid_aws_account_id,
+        'help': """
+                ID of the AWS account, where AWS CloudFormation stacks will be
+                deployed so that CI artifacts can be viewed online.
+                This option should be used only by public GitHub repositories.
+                """
         }, {
         'flag': '--cbmc',
         'metavar': '<X.Y.Z>',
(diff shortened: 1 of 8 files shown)


src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
```diff
@@ -116,4 +147,19 @@ def print_proof_results(out_file):
 
 if __name__ == '__main__':
     args = get_args()
-    print_proof_results(args.run_file)
+    with_aws = args.cloudfront_domain and args.s3_uri
+    without_aws = not args.cloudfront_domain and not args.s3_uri
+    if not with_aws and not without_aws:
+        raise argparse.ArgumentTypeError(
```
karkhaz, line 153: I don't think this is the right thing to do, this will dump a stack trace. `argparse` has [`exit()` and `error()`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error) methods that print out the generated help text, plus your message to stderr and exit with 2, that's probably what you want here

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
```diff
@@ -116,4 +147,19 @@ def print_proof_results(out_file):
 
 if __name__ == '__main__':
     args = get_args()
-    print_proof_results(args.run_file)
+    with_aws = args.cloudfront_domain and args.s3_uri
+    without_aws = not args.cloudfront_domain and not args.s3_uri
+    if not with_aws and not without_aws:
```
karkhaz, line 152: Nit (De Morgan's law): I think `if not (with_aws or without_aws)`, or TBH `if not (both_args or neither_arg)` reads more naturally

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
```diff
@@ -116,4 +147,19 @@ def print_proof_results(out_file):
...
+        raise argparse.ArgumentTypeError(
+            "The CloudFront domain and the S3 URI should either be specified "
+            "together or they should not be specified altogether.")
+    if with_aws:
+        CBMC_PROOF_REPORT_HEADER = "Click here to see the CBMC proof report"
```
karkhaz, line 157: why capitals?

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
```diff
@@ -116,4 +147,19 @@ def print_proof_results(out_file):
...
 if __name__ == '__main__':
     args = get_args()
-    print_proof_results(args.run_file)
+    with_aws = args.cloudfront_domain and args.s3_uri
+    without_aws = not args.cloudfront_domain and not args.s3_uri
+    if not with_aws and not without_aws:
+        raise argparse.ArgumentTypeError(
+            "The CloudFront domain and the S3 URI should either be specified "
+            "together or they should not be specified altogether.")
+    if with_aws:
```
karkhaz, line 156:
```
if without_aws:
  print_proof_results(args.run_file)
  return 0
```
then dedent the rest

src/cbmc_starter_kit/util.py
```diff
@@ -88,6 +115,21 @@ def ask_for_function_name():
 
     return input("What is the function name? ").strip()
 
+def ask_for_aws_account():
```
karkhaz, line 118: This is not used?

karkhaz (changes requested):

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
karkhaz: I think it's okay to raise `ArgumentTypeError` inside an argparse 'type' function, because argparse specifically catches that exception while processing the argument type. But it's not meant to be raised from arbitrary places in the code I think

src/cbmc_starter_kit/setup_ci.py
```diff
@@ -42,6 +48,14 @@ def parse_arguments():
                 run CBMC proofs in GitHub Actions. If your repo has a large
                 runner available, you can specify it here. default: %(default)s"""
         }, {
+        'flag': '--aws-account-id',
+        'type': lambda x: x if x.isdigit() and len(x) == 12 else False,
```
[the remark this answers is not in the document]
angelonakos: > it's basically going to ignore the user's input if it isn't a number (because you check if args.aws_account_id below). Maybe I don't understand but I think this is a bit confusing? If the user gives an invalid command line arg the usual behavior is to exit?

If the user doesn't provide a value that is a 12-digit string, then the command will exit.

> Why not write a proper function that raises ValueError so that the argparser can handle it and print an error message?

You don't have to. The use of an anonymous function here is actually the most concise option in terms of handing invalid user input for this argument.

src/cbmc_starter_kit/setup_ci.py
```diff
@@ -77,6 +91,131 @@ def parse_arguments():
...
+            logging.info("Creating %s", stack_name)
+            stack_result = cf_client.create_stack(**params)
+            waiter = cf_client.get_waiter("stack_create_complete")
+        logging.info("Waiting for stack operation to complete...")
+        waiter.wait(StackName=stack_name)
+    except botocore.exceptions.ClientError as ex:
+        error_message = ex.response["Error"]["Message"]
+        if error_message != "No updates are to be performed.":
+            logging.error(
+                "When updating stack, unexpected error encountered: %s",
```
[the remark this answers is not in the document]
angelonakos: > log the actual error message? 

Updated.

> or does boto3 already print it to stderr?

yes

src/cbmc_starter_kit/setup_ci.py
```diff
@@ -77,6 +91,131 @@ def parse_arguments():
...
+                stack_name,
+            )
+            raise
+        print(f"No updates are to be performed for stack {stack_name}")
+    else:
+        logging.info(json.dumps(stack_result, indent=4))
```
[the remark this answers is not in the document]
angelonakos: > could you at least write a prefix like logging.info("Stack '%s' deployment result: %s", stack_name, stack_result) or something?

Added a new log statement to indicate the name of the stack, such that reader knows for which stack the "deployment result" pertains to.

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
angelonakos: Using first one

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
angelonakos: Because it is a constant and easier to **not** skip when a future reader goes over this code and wants to update the heading.

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
angelonakos: Must `s/return 0/exit(0)` because this is not inside a function. But, yes, I will use this.

src/cbmc_starter_kit/util.py
angelonakos: Will remove.

src/cbmc_starter_kit/template-for-repository/proofs/lib/summarize.py
angelonakos: I personally prefer having both of these arguments in a subparser to make the condition that both should be called/not called in unison more obvious. However, after earlier offline chat, it was decided that a subparser wouldn't be used here

angelonakos: I will be using `error()` as you pointed out.

src/cbmc_starter_kit/setup_ci.py
```diff
@@ -77,6 +91,131 @@ def parse_arguments():
...
+        "secrets used in GitHub Actions.\nThe secret must have this value:\n")
+    response = cf_client.describe_stacks(StackName=pipeline_stack)
+    outputs = response["Stacks"][0]["Outputs"]
+    aws_cloudfront_domain = "''"
+    for output in outputs:
+        key_name = output["OutputKey"]
+        if key_name == "ProofActionsRoleArn":
+            print(output["OutputValue"])
+        elif key_name == "ProofCIDistributionDomainName":
+            aws_cloudfront_domain = output["OutputValue"]
```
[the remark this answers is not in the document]
angelonakos: I was using this approach as it's unclear from documentation whether the keys always come in the same order or not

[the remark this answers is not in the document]
angelonakos: If I `return output["OutputValue"]` here, I might end up not having printed the ARN of the IAM role (assuming that key was next in the list of outputs)

src/cbmc_starter_kit/setup_ci.py
```diff
@@ -141,19 +281,36 @@ def main():
...
+    cfn_path = workflows_root / "proof_ci_resources" / "cfn"
     workflow = workflows_root / "proof_ci.yaml"
 
     shutil.copytree(
         util.package_ci_workflow_template_root(),
         workflows_root,
         dirs_exist_ok=True)
 
-    patch_proof_ci_config(config, args)
+    aws_cloudfront_domain = "''"
```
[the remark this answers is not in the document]
angelonakos: > Why are we using the special string "''"?

If you look at the `proof_ci.yaml` file, you will see that certain steps are executed based on whether or not `AWS_CLOUDFRONT_DOMAIN` is equal to an empty string (`''`) or not.

src/cbmc_starter_kit/setup_ci.py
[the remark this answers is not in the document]
angelonakos: > Does the user need to see it at all? 

Ronak pointed out to me that it makes sense to log this "stack deployment result" considering that the code earlier logs the name of the stack being created/updated. I agree with that sentiment.

src/cbmc_starter_kit/setup_ci.py
```diff
@@ -77,6 +91,131 @@ def parse_arguments():
...
+                {"Key": "repo", "Value": repo_name}
+            ],
+        },
+    )
+    print(
+        "Visit:\n\n"
+        f"https://github.com/{repo_owner}/{repo_name}/settings/secrets/actions\n")
+    print(
+        "and add a \"PROOF_CI_IAM_ROLE\" secret to your GitHub repository's "
+        "secrets used in GitHub Actions.\nThe secret must have this value:\n")
```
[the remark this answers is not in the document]
angelonakos: > I would do the for loop on line 180 first, save the role arn to a variable, and then print the whole message if it was successfully found, and exit if not. 

Will do.

src/cbmc_starter_kit/setup_ci.py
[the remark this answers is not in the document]
angelonakos: > deploy() should return None IMHO.

Good point. I split out the relevant code into smaller functions.

src/cbmc_starter_kit/setup_ci.py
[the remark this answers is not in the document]
angelonakos: I was wrong. I have updated the `type` to be a function that will raise ArgumentTypeError if the provided input is not a 12-character string consisting only of digits.

karkhaz:
Thank you for implementing this, I think we can keep the branch for historical reasons. Closing as we won't be supporting this feature.

PR #170, opened 2022-12-21
Update workflows

markrtuttle:
Update workflows to latest github action version numbers, update github action used to create release

Also update version to 2.8.7

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -12,7 +12,7 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.RELEASE_CI_ACCESS_TOKEN }}
     steps:
       - name: Checkout code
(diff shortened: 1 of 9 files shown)


.github/workflows/release.yaml
```diff
@@ -26,32 +26,6 @@ jobs:
             exit 1
           fi
       - name: Create release
-        uses: actions/create-release@v1
+        uses: ncipollo/release-action@v1
```
ronakfof, line 29: This is a third party GitHub Action. Are you sure you want to use it?

markrtuttle: It recommended by github:  See "maintained actions" on https://github.com/actions/create-release

ronakfof (approved):

Issue #173, opened 2023-01-06
Add versioning to ProofCI GitHub Actions workflow file


angelonakos:
@kstribrnAmzn asked during a demo of the ProofCI GitHub Actions whether the workflow file is versioned.  It currently isn't.

We should **add** a comment within [proof_ci.yaml](https://github.com/model-checking/cbmc-starter-kit/blob/master/src/cbmc_starter_kit/template-for-ci-workflow/proof_ci.yaml) that indicates the version of the CBMC starter-kit that generated/contributed this workflow file à la Makefile.common

kstribrnAmzn:
Thanks for tagging me here! I'll keep my eyes on this issue as it progresses.

angelonakos:
Resolved.

PR #174, opened 2023-01-09
Version GitHub Actions workflow file

angelonakos:
*Issue #, if available:*

#173 

*Description of changes:*

If a target proof-project repository lacks the GitHub Actions workflow file, when the user runs the `cbmc-starter-kit-setup-ci` the comment in the template will be updated to reflect the version of the CBMC starter kit that "introduces" the workflow file to a target repository.

If the target proof-project already contains the GitHub Actions file to run CBMC proofs, then the comment indicating the version of the CBMC starter kit is being appropriately updated.

For example, assume repository X has version X.Y.Z mentioned near the top of the ProofCI workflow file and that version A.B happens to be the latest semantic version of the CBMC starter kit. During a CBMC upgrade, the comment in the workflow file will be updated to reflect version A.B

Testing:

If I change the version in `setup.cfg` and `src/cbmc_starter_kit/version.py` to 2.8.11 and run `cbmc-starter-kit-setup-ci` in a target proof project, the resulting workflow file for the ProofCI GitHub Action will contain a comment in the top indicating the version - 2.8.11 in thie case - of the CBMC starter kit that was used to introduce the workflow file. If I change the version to 2.23 and re-build the package and attempt to update a proof-project that contains the workflow file, then the version will change from 2.8.11 to 2.23

### DIFF
--- a/src/cbmc_starter_kit/version.py
+++ b/src/cbmc_starter_kit/version.py
@@ -3,6 +3,8 @@
 
 """Version number."""
 
+import os
+
 NAME = "CBMC starter kit"
 NUMBER = "2.8.8"
 VERSION = f"{NAME} {NUMBER}"
@@ -21,3 +23,13 @@ def copy_with_version(src, dst):
         data = srcfile.read()
     with open(dst, 'w', encoding='utf-8') as dstfile:
         dstfile.write(data.replace(REPLACE_TARGET, version()))
+
+def update_existing_version_in_workflow_file(workflow):
+    tmp_file = f"{workflow}~"
+    with open(workflow) as src, open(tmp_file, "w") as dst:
+        for line in src:
+            if line.startswith(f"# {NAME}"):
+                print(version(), file=dst)
+            else:
+                print(line, file=dst)
+    os.rename(tmp_file, workflow)
--- a/src/cbmc_starter_kit/setup_ci.py
+++ b/src/cbmc_starter_kit/setup_ci.py
@@ -10,7 +10,7 @@ import shutil
 from argparse import RawDescriptionHelpFormatter
 
 
-from cbmc_starter_kit import arguments, repository, util
+from cbmc_starter_kit import arguments, repository, util, version
 
 ################################################################
 
@@ -153,6 +153,7 @@ def main():
 
     patch_proof_ci_config(config, args)
     patch_proof_ci_workflow(workflow, args.github_actions_runner)
+    version.copy_with_version(workflow, workflow)
 
     repo_root = repository.repository_root()
     copy_lib_directory(repo_root)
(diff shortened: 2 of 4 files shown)


src/cbmc_starter_kit/version.py
```diff
@@ -21,3 +21,15 @@ def copy_with_version(src, dst):
         data = srcfile.read()
     with open(dst, 'w', encoding='utf-8') as dstfile:
         dstfile.write(data.replace(REPLACE_TARGET, version()))
+
+def update_existing_version_in_workflow_file(workflow):
+    with open(workflow, encoding='utf-8') as srcfile:
```
karkhaz, line 26: why is `encoding` needed?

src/cbmc_starter_kit/version.py
```diff
@@ -21,3 +21,15 @@ def copy_with_version(src, dst):
...
+        lines = srcfile.read().splitlines()
+    new_lines = []
+    for line in lines:
```
karkhaz, line 29: You can actually iterate over `srcfile`, like this
```python
with open("..") as srcfile:
  for line in srcfile:
        if line.startswith(f"# {NAME}"):
            new_lines.append(version())
```
It's better to process lines from a file one at a time rather than loading them all into a buffer if possible because it saves memory. It probably doesn't matter here but there's no reason not to set a good example.

In this case I don't think you even need an output buffer, I would recommend this:
```python
tmp_file = f"{workflow}~"
with open(workflow) as src, open(tmp_file, "w") as dst:
    for line in src:
        if line.startswith(f"# {NAME}"):
            print(version(), file=dst)
        else:
            print(line, file=dst)
os.rename(tmp_file, workflow)
```

karkhaz (changes requested):

src/cbmc_starter_kit/version.py
angelonakos: This was a great and concise suggestion. I implemented it and it worked. I tested it by running `cbmc-starter-kit-setup-ci` in these 2 branches:
- https://github.com/ronakfof/private-coreMQTT-Agent/tree/add-gha-workflow-test-version (this branch tests that an existing file is updated and the version is inserted appropriately)
- https://github.com/ronakfof/private-coreMQTT-Agent/tree/test-version (this branch tests that a new GitHub Actions ProofCI workflow file will appropriately display the version of the CBMC starter kit)

src/cbmc_starter_kit/version.py
angelonakos: Removed. Indeed extraneous for this case.

karkhaz (approved):

karkhaz:
Thank you!

PR #176, opened 2023-01-11
Use GOTO-style numbering for goto binaries

karkhaz:
Prior to this commit, adding a new instrumentation to the starter kit required renaming all subsequent goto binaries after the instrumentation. This results in confusing diffs that are difficult to review. This commit changes the naming scheme so that new instrumentations can be added between existing ones.


tautschnig (approved):
Thank you!!

PR #195, opened 2023-03-13
Only unwind loops without contracts in DFCC mode

feliperodri:
*Description of changes:*

We should **not** early unroll loops using `--unwind 1` if loop contracts are presented in DFCC mode. This change captures this case.


src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -711,15 +711,27 @@ endif
 # Early unwind all loops on DFCC mode; otherwise, only unwind loops in proof and project code
 $(HARNESS_GOTO)0500.goto: $(HARNESS_GOTO)0400.goto
 ifneq ($(strip $(USE_DYNAMIC_FRAMES)),)
+ifneq ($(strip $(APPLY_LOOP_CONTRACTS)),)
```
karkhaz, line 714: This is a lot of lines of code for effectively two changes: adding `$(CBMC_DEFAULT_UNWIND)` and changing the description.

Would you consider a different approach: near the top of the file, do something like

```
ifneq ($(strip $(USE_DYNAMIC_FRAMES)),)
ifneq ($(strip $(APPLY_LOOP_CONTRACTS)),)
UNWIND_0500_FLAGS=$(CBMC_FLAG_UNWINDING_ASSERTIONS)
UNWIND_0500_DESC=
else
UNWIND_0500_FLAGS=$(CBMC_FLAG_UNWINDING_ASSERTIONS) $(CBMC_DEFAULT_UNWIND)
UNWIND_0500_DESC="$(PROOF_UID): unwinding all loops"
endif
```

then down here, you don't need to branch, just use the value of `UNWIND_0500_FLAG` in the command.

karkhaz: I'm not going to block on this. It seems easier to to me to read the recipe this way, and if the reader cares about what the flags are they can check.

On the other hand it is less explicit and does not call attention to the fact that the flags might change depending on the proof.

karkhaz (approved):

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
feliperodri: Great catch, Kareem! We should avoid code duplication is not less clear than how it was before.

karkhaz:
Thank you! 🚢

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -450,6 +450,23 @@ ifdef APPLY_LOOP_CONTRACTS
   endif
 endif
 
+# The default unwind should only be used in DFCC mode without loop contracts.
+# When loop contracts are applied, we only unwind specified loops.
+# If any loops remain after loop contracts have been applied, CBMC might try
+# to unwind the program indefinetly, because we do not pass default unwind
+# (i.e., --unwind 1) to CBMC when in DFCC mode.
+# We must not use default unwind with DFCC mode, because loops introduced during
+# contract instrumentation must be dynamically unwind during symex.
```
remi-delmas-3000, line 459:
```suggestion
# We must not use a default unwind command in DFCC mode, because contract instrumentation
# introduces loops encoding write set inclusion checks that must be dynamically unwound during
# symex.
```

src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -450,6 +450,23 @@ ifdef APPLY_LOOP_CONTRACTS
...
+ifneq ($(strip $(USE_DYNAMIC_FRAMES)),)
+ifneq ($(strip $(APPLY_LOOP_CONTRACTS)),)
+  UNWIND_0500_FLAGS=$(CBMC_UNWINDSET) $(CBMC_CPROVER_LIBRARY_UNWINDSET) $(CBMC_FLAG_UNWINDING_ASSERTIONS)
+  UNWIND_0500_DESC="$(PROOF_UID): unwinding all loops without loop contracts"
```
remi-delmas-3000, line 463:
```suggestion
  UNWIND_0500_DESC="$(PROOF_UID): unwinding specified subset of loops"
```

remi-delmas-3000 (changes requested):

remi-delmas-3000 (approved):

PR #224, opened 2025-01-23
Add new optional flags for the cprover contracts library.

remi-delmas-3000:
*Description of changes:*
-  defining `DFCC_DEBUG_LIB` in proof makefiles activates debug asserts in cprover_contracts.c
- defining `DFCC_SIMPLE_INVALID_POINTER_MODEL` in proof makefiles removes offset nondeterminism for invalid pointers produced in failure paths of the `__CPROVER_is_fresh` predicate. Mitigates some proof performance issues, but is possibly unsound due to reduced nondeterminism.

### DIFF
--- a/src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
+++ b/src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
@@ -407,9 +407,29 @@ ifdef STRING_ABSTRACTION
   endif
 endif
 
+# User simpler invalid pointer model for contracts with DFCC.
+#
+# Removes offset nondeterminism for invalid pointers created in failure paths
+# of the __CPROVER_is_fresh predicate. Improves performance for some proofs.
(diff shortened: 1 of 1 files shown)


src/cbmc_starter_kit/template-for-repository/proofs/Makefile.common
```diff
@@ -407,9 +407,29 @@ ifdef STRING_ABSTRACTION
   endif
 endif
 
+# User simpler invalid pointer model for contracts with DFCC.
+#
+# Removes offset nondeterminism for invalid pointers created in failure paths
+# of the __CPROVER_is_fresh predicate. Improves performance for some proofs.
+# Unsound because of reduced nondeterminism.
```
feliperodri, line 414: Is there a way that we make this more clear to users? Maybe add unsound to the name of the flag? Does CBMC emit warnings when this flag is used? Is there a PR or GitHub issue link that we could add here?

